PR #6, opened 2022-04-08
Initial Public FS and WebAssembly Work

### LINKED ISSUE #4: Initial IPLD work and Wasm-bindgen Setup
Implement a basic WNFSv2 public filesystem as a proof of concept of running wasm in the browser. This is not supposed to be a full implementation of the public filesystem. 

To resolve this issue, the following conditions should met:

- separating core library from wasm-specific code
- asynchronous function call from JavaScript
- a demo to show the pieces working together in the web

appcypher:
## Summary

* [x] Initial Public FS implementation that immutable core, filetree merging, etc. will be based on.
* [x] Rust unit tests 
* [x] A wasm-bindgen interface implementation.

This is the first of a few PRs that will implement the WNFSv2 Public file system.

## Test plan (required)

-  Testing the Rust core.
    
    ```bash
    cargo test -p wnfs
    ```

- Testing the wasm binding.
  
  ```bash
  cd crates/wasm
  ```
  
  ```bash
  wasm-pack test --chrome
  ```
  
  ```bash
  open http://127.0.0.1:8000
  ```

## Closing issues

Fixes #4  

## Ongoing Issues
- https://github.com/fission-suite/webnative/issues/371


crates/fs/public/directory.rs
```diff
@@ -0,0 +1,901 @@
...
+        Ok(OpResult {
+            root_node,
+            result: working_node,
+            diverged: diverge,
+        })
+    }
+
+    /// Looks up a node by its path name in the current directory.
+    ///
+    /// TODO(appcypher): What is a valid path segment identifier?
```
matheus23, line 148: Yeah, that's a great question. I can only link an old internal discussion about this: https://talk.fission.codes/t/valid-paths-in-wnfs/2015

My thinking is, let's assume UTF8. Technically any byte array is valid (even ones that utf8-decoded include a slash), because of how the encoding into IPLD works out.

This is something we'd need to think about a little harder if we want to implement the pretty tree maybe.

crates/fs/public/directory.rs
```diff
@@ -0,0 +1,901 @@
...
+/// Represents a directory that has possibly diverged. It is the result of operating on a directory.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OpResult<T> {
+    // The root node. It is the same as the previous root node if the directory has not been diverged.
+    pub root_node: Shared<PublicNode>,
+    // Implementation dependent but it usually the last leaf node operated on.
+    pub result: T,
+    /// Whether this is a divergence or not.
+    pub diverged: bool,
+}
```
matheus23, line 48: I think we might be able to go without the `OpResult` type.
* `get_node` returns an `OpResult`, but that means you get back a `PublicNode` in `root_node` and `result`, where I'd pretty much expect `root_node` to be the same thing that that's used to call `get_node` on.
* `read` returns an `OpResult`, but like `get_node`, it returns back the `root_node` you originally called `read` on.
* `write` returns the parent of the written node. That `result` isn't used anywhere at the moment.

`diverged` may not be needed when stuff is immutable, but I may just not really understand it yet! :sweat_smile:

crates/fs/common/blockstore.rs
```diff
@@ -0,0 +1,137 @@
...
+#[async_trait(?Send)]
+pub trait BlockStoreLookup {
+    async fn get_block<'a>(&'a self, cid: &Cid) -> Result<Cow<'a, Vec<u8>>>;
+}
+
+/// For types that implement loading decodable object from a blockstore using a CID.
+#[async_trait(?Send)]
+pub trait BlockStoreCidLoad {
+    /// Loads a decodable object from the store with provided CID.
+    async fn load<T: Decode<C>, C: Codec>(&self, cid: &Cid, decoder: C) -> Result<T>;
```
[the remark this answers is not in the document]
appcypher: You are right. We don't need to make it a trait here.

matheus23 (approved):
First end-to-end public filesystem implementation! :tada: 

dance

Stephen and I talked about the comments I left above: The crux of it would be a change somewhat deeper down & technically only a refactor, the surface API would stay the same.
So let's capture that refactor (figuring out how/whether we can remove the `RefCell`, and make all algorithms only make use of mutation locally without needing a refcell and non-recursive) in an issue and we'll tackle that as the next PR. :+1: 
Technically we can continue poking the wasm blob this generates :smile:

PR #18, opened 2022-05-13
Try to remove RefCell & try PathTo abstraction

matheus23:
Co-authored-by: appcypher <appcypher@users.noreply.github.com>

## Closing issues

Fixes https://github.com/WebNativeFileSystem/rs-wnfs/issues/10

### DIFF
--- a/crates/fs/public/directory.rs
+++ b/crates/fs/public/directory.rs
@@ -28,11 +28,8 @@ use super::{DeepClone, Id, Link, PublicFile, PublicNode};
 //--------------------------------------------------------------------------------------------------
 
 /// A directory in a WNFS public file system.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct PublicDirectory(Shared<PublicDirectoryInner>);
-
 #[derive(Debug, Clone, PartialEq, Eq, FieldNames)]
-struct PublicDirectoryInner {
+pub struct PublicDirectory {
     metadata: Metadata,
     userland: BTreeMap<String, Link>,
     previous: Option<Cid>,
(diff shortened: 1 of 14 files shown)


matheus23:
@appcypher I continued work on this a little bit, I'm fairly confident we could try to remove our dependency on `RefCell` like this :)
Let me know what you think of this.
One thing it's doing differently is using `PathTo` in cases where you'd use `upsert` before. Think of it like `get_node_path` and `get_node_path_with_mkdir` abstracting away everything that would happen *before* the Fn in `upsert` is called, and `reconstruct_dir` abstracting what would happen *after* it.

matheus23:
@appcypher and I paired on this, we've got some more goodies into this. :raised_hands: 

Remaining work is
* Rebasing this branch on main
* Modifying `baseHistoryOn` to work on the `Rc` stuff instead of the `Rc<RefCell` stuff

crates/fs/public/directory.rs
```diff
@@ -538,10 +537,10 @@ impl PublicDirectory {
     }
 
     /// Gets the iterator for walking the history of a directory node.
-    pub fn get_history<'a, B: BlockStore>(
+    pub fn get_history<B: BlockStore>(
         self: Rc<Self>,
-        store: &'a B,
-    ) -> impl Stream<Item = Result<Cid>> + 'a {
+        store: &B,
+    ) -> impl Stream<Item = Result<Cid>> + '_ {
```
matheus23, line 543: Huh interesting. So that's the same as saying
```rust
pub fn get_history<'a, B: BlockStore>(
    self: Rc<Self>,
    store: &B,
) -> impl Stream<Item = Result<Cid>> + 'a
```
right?

I'm not quite sure what exactly is happening here. Why do we have to annotate this lifetime, it's not like we return something borrowed, right? The caller will own the stream, no?

(This is me learning rust again, it's not about changing our code here :D )

crates/wasm/fs/public/directory.rs
```diff
@@ -152,7 +157,7 @@ impl PublicDirectory {
         time: &js_sys::Date,
```
matheus23, line 157: Unfortunately I can't comment on line 148 because it wasn't changed in this PR. But why do we have to pass a `&mut self`. We're not really mutating the `PublicDirectory`, so I'm not sure why it would be needed :thinking:

matheus23 (commented):
LOL apparently I can't approve my own PR in github. Fair enough I guess :laughing: 

I'm giving this a :+1: :heavy_check_mark: this way then I guess :stuck_out_tongue: 

I love the renames, we really needed to clean that up! Thanks :pray: 

I just have one learning Q about lifetimes and another Q regarding usage of `self: &mut PublicDirectory` in the wasm crate. But we can go ahead and merge before these Qs are answered :)

matheus23:
You'll have to approve this PR yourself if you want it to get merged @appcypher :laughing: 

I don't know why I think that's so funny

crates/wasm/fs/public/directory.rs
appcypher: You are right. They are residues of refactoring. Removed.

crates/fs/public/directory.rs
appcypher: The `try_stream!` macro creates a Stream object that gets returned in this function. And Streams, like Futures and Closures, can capture references from outer scope. In this case the Stream is borrowing the `store` object and the compiler has to make sure `store` lifetime exceeds that of the returned Stream object, hence the annotation on the return type.

Because writing lifetimes can be unwieldy sometimes, Rust introduced lifetime elision rules, which are some of the common cases where the compiler is able to infer a lifetime so you don't have to be explicit about it. You can read up on the rules for elision here. https://doc.rust-lang.org/nomicon/lifetime-elision.html

In this case, there is only one input lifetime (`store: &B`), which automatically propagates to all output lifetimes, following the second rule in that book.

Even though we need to specify that there is a lifetime on the output, we don't need to specify which lifetime.

PR #21, opened 2022-05-20
Publish Crates, Npm Package and Usage Docs

appcypher:
## Summary

* [x] Publish new crates.
* [x] Publish to npm.
* [x] Add usage examples

This PR just prepares the project for package publishing (npm, doc.rs, crates.io). Also adds usage examples.

**WNFS NPM package:**
- https://www.npmjs.com/package/wnfs

**WNFS crates:**
- https://crates.io/crates/wnfs
- https://crates.io/crates/wasm-wnfs

**Examples of WNFS usage here:**
JavaScript:
- https://replit.com/@appcypher/wnfs-npm-test#index.js

Rust:
- https://replit.com/@appcypher/wnfs-crate-test#src/main.rs
## Test plan (required)

-  `N/A`

## Ongoing Issues
- https://github.com/fission-suite/webnative/issues/371

## Closing issues

Fixes https://github.com/WebNativeFileSystem/rs-wnfs/issues/22


README.md
```diff
@@ -85,9 +180,9 @@ Its goal is to be as dependency-less as possible in order to be easily compiled
 
   </details>
 
-- **The _wnfs_ Helper Script**
+- **The _wnfs_ Command**
```
matheus23, line 183: Maybe we should rename the script in order to avoid confusion with a potential wnfs CLI that actually reads and writes WNFS similar to what the `ipfs` CLI is doing today.

Or we could go for a `Justfile` which is similar to a makefile, but simpler. We're using it in a lot of other places at fission.

And lastly we could go for a nix version, but that's way more involved to set up compared to a justfile.

matheus23 (approved):
Just a comment about the `wnfs` scripts (and generally some ideas around where to go from there with out builds scripts).
I think we should at least rename it, but other than that this is good to go!

Docs :raised_hands: :sparkles: 

:shipit:

README.md
appcypher: I was considering make but had issues with it in the past. Just looks like an improved make. I'm going to switch to it later.
I agree with renaming it. Maybe `rs-wnfs` because it  specific to the repo?

PR #28, opened 2022-06-06
Initial Private FileSystem

### LINKED ISSUE #14: TRACKING: Private FS Work
Implement WNFSv2 private filesystem. This will be one of the initial work to get WNFSv2 private filesystem working.
Here are some of the things that should be implemented:

- [x] HAMT for file tree storage (https://github.com/WebNativeFileSystem/rs-wnfs/issues/24)
- [x] Bloom/name filter with ancestry check support (https://github.com/WebNativeFileSystem/rs-wnfs/issues/26)
- [x] Skip ratchet for backward secrecy and forward transparency (https://github.com/WebNativeFileSystem/rs-skip-ratchet/issues/1)
- [x] IPLD serialization (https://github.com/WebNativeFileSystem/rs-wnfs/issues/29)
- [x] Private filesystem APIs (Rust and WebAssembly) (https://github.com/WebNativeFileSystem/rs-wnfs/issues/30)
- [x] Encrypt Header Inside the Private Node (https://github.com/wnfs-wg/rs-wnfs/issues/47)
- [x] Implement Seeking & Seeking + Recovering Reads - (https://github.com/wnfs-wg/rs-wnfs/issues/46)
- [x] Implement `mv` (https://github.com/wnfs-wg/rs-wnfs/issues/54)

### LINKED ISSUE #27: Use `OnceCell` to cache decoded CID links
We're thinking of a data structure that may look like this:

```rust
enum Link<T> {
    Clean {
        cid: Cid,
        cache: OnceCell<Box<T>>,
    },
    Dirty(Box<T>),
}
```

There may be issues using a `OnceCell`, we need to figure it out!
Also, to `Box` or not to `Box` (or `Rc`?) TBD.

appcypher:
## Summary

Implement a basic WNFSv2 private filesystem. This will be one of the initial work to get WNFSv2 private filesystem working.

- [x] HAMT and Node Data Structure
- [x] IPLD Serialisation

Explain the **motivation** for making this change. What existing problem does the pull request solve?

## Test plan (required)
-  Testing the Rust core.
    
    ```bash
    cargo test -p wnfs --release
    ```

- Testing the wasm bindings.
  
  ```bash
  cd crates/wasm
  ```
  
  ```bash
  yarn playwright test
  ```

## Closing issues

Fixes #27


appcypher:
Serde Visitor implementation for enums that use Deserialize derive doesn't work well with libipld 128-bit integer format.
https://github.com/serde-rs/serde/issues/2230

Tried to fix it with a fork but it only broke more deps down the line. I managed to fix all of them but it is a very hacky unpleasant solution, so I'm working on an alternative solution that won't be as intrusive with custom Visitor impl.

I think we still need serde because a good chunk of rust ecosystem including libipld which we rely on atm, leverage it. Implementing sth similar to serde is a non-goal.

crates/fs/common/metadata.rs
```diff
@@ -69,7 +61,7 @@ pub struct UnixFsMetadata {
 }
 
 /// The metadata of a node on the WNFS file system.
-#[derive(Debug, Clone, PartialEq, Eq, FieldNames)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
matheus23, line 64: Do we still need `derive(Deserialize)` on `Metadata`, `UnixFsMetadata` and `UnixFsMode` now that we implemented `TryFrom<&Ipld>` for them?

It may be confusing since there's two ways to serde-deserialize them then: serde -> Ipld -> Metadata and serde -> Metadata directly.

crates/fs/private/hamt/hash.rs
```diff
@@ -0,0 +1,100 @@
...
+// Type Definition
+//--------------------------------------------------------------------------------------------------
+
+pub type HashOutput = [u8; HASH_BYTES];
+
+pub trait GenerateHash {
+    fn generate_hash<K: AsRef<[u8]>>(key: &K) -> HashOutput; // &[u8]
+}
+
+pub struct HashQuartets<'a> {
```
matheus23, line 27: Googling around, it looks like quartet was used for 4-bit segments in telecommunications stuff. I didn't know that term before!
Maybe it's just me, but there's some more common terms like "Nibble" "half-byte" or "octet". Would they work better?

crates/fs/private/hamt/node.rs
```diff
@@ -0,0 +1,251 @@
...
+        self: &'a Rc<Self>,
+        key: &K,
+        store: &B,
+    ) -> Result<Option<&'a V>> {
+        Ok(self.search(key, store).await?.map(|pair| &pair.value))
+    }
+
+    /// Recursively resolves all nodes under the directory to Cids.
+    #[async_recursion(?Send)]
+    pub async fn flush<'a, B: BlockStore>(self: &'a Rc<Self>, store: &mut B) -> Result<()> {
```
matheus23, line 64: We should be able to implement `AsyncSerialize` for the HAMT nodes now instead :+1:

crates/fs/private/hamt/node.rs
```diff
@@ -0,0 +1,251 @@
...
+impl<K, V> Serialize for Node<K, V>
+where
+    K: Serialize,
+    V: Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        (&self.bitmask.as_raw_slice(), &self.pointers).serialize(serializer)
```
matheus23, line 204: Wait, how is that even possible? I don't see a `impl Serialize for Pointers` in `pointers.rs`? :thinking:

crates/fs/public/directory.rs
```diff
@@ -425,16 +429,17 @@ impl PublicDirectory {
     /// }
     /// ```
     pub async fn store<B: BlockStore>(&self, store: &mut B) -> Result<Cid> {
-        let bytes = self.encode(store).await?;
-        store.put_block(bytes, IpldCodec::DagCbor).await
+        store.put_async_serializable(self).await
+        // self.flush(store).await?;
+        // store.put_serializable(self).await
```
matheus23, line 434: I know this code is probably WIP. I'm making sure we don't forget to remove the commented lines :sweat_smile:

crates/fs/public/directory.rs
```diff
@@ -237,6 +236,11 @@ impl PublicDirectory {
         }
     }
 
+    /// Creates a new directory using the given metadata.
+    pub fn with_rc(time: DateTime<Utc>) -> Rc<Self> {
+        Rc::new(Self::new(time))
+    }
+
```
matheus23, line 243: Sorry, I may sound harsh, but do we need this? This saves ~5 characters on call sites compared to `Rc::new(PublicDirectory::new(...))` plus a set of parenthesis. As far as I can tell we only use it once.

Or is `with_rc` possibly a common idiom in Rust that I don't know about?

crates/fs/public/file.rs
```diff
@@ -78,13 +77,9 @@ impl PublicFile {
...
     /// ```
+    #[inline(always)]
     pub async fn store<B: BlockStore>(&self, store: &mut B) -> Result<Cid> {
-        let bytes = {
-            let mut tmp = vec![];
-            self.encode(DagCborCodec, &mut tmp)?;
-            tmp
-        };
-        store.put_block(bytes, IpldCodec::DagCbor).await
+        store.put_serializable(self).await
```
matheus23, line 82: Now that this has become so much simpler, I'd vote for just having `store.put_serializable(file)` instead of both that and `file.store(store)`. I think it's important to keep the choices small, so it's clear what one should use in each situation.

crates/fs/public/node.rs
```diff
@@ -1,36 +1,37 @@
...
+//--------------------------------------------------------------------------------------------------
+// Type Definitions
+//--------------------------------------------------------------------------------------------------
 
 /// A node in a WNFS public file system. This can either be a file or a directory.
-#[derive(Debug, Clone, PartialEq, Eq)]
+///
+/// PublicNode is serialized and deserialized as [untagged][1] enum.
+///
+/// [1]: https://serde.rs/enum-representations.html#untagged
```
matheus23, line 22: Again, I know, WIP, but just making sure we don't forget to remove this :)

crates/fs/traits.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// Implements getting a unique identifier for a node.
+pub trait Id {
+    /// Gets an identifier for the node.
+    fn get_id(&self) -> String;
+}
+
+/// Implements deep equality check for two types.
+#[async_trait(?Send)]
+pub trait DeepEq {
```
matheus23, line 20: Technically `Eq` is already supposed to be a sort of deep equality (I think).
The biggest difference is that this is deep equality for IPLD. So maybe we call this `IpldEq`? Or `LinkedDataEq` or something like that? To a reader that may explain better why you need to provide it with a blockstore.

matheus23 (commented):
I have a bunch of comments. Just small stuff :)

I'm excited we've found a fairly good serialization/deserialization strategy :+1:

crates/fs/private/hamt/hash.rs
appcypher: Let's do nibble

crates/fs/public/node.rs
appcypher: I need to reword this because it is still serialized untagged

crates/fs/public/directory.rs
appcypher: Let's get rid of it. I thought I would need it a lot.

matheus23:
Closes #29

crates/fs/common/metadata.rs
appcypher: We do actually. Ipld intermediate does not make them work with serde. Also it is used in PublicFile and PublicDirectory.

crates/fs/public/file.rs
appcypher: `store` is exposed as a public API. `Blockstore::put_serializable` isn't, so we need to have it.

appcypher: The issue there is we cannot expose a generic interface from WebAssembly.

matheus23 (approved):
Let's ship it! :shipit: :tada:

PR #31, opened 2022-06-28
Namefilter and HAMT Implementation and Tests

appcypher:
## Summary

Implement the updated HAMT and Namefilter specs. The new spec updates the the saturation and hash algorithm used by the namefilter. Some decisions were also made around what kind of degree WNFS HAMT should have among other things.

- [x] Updated WNFS HAMT spec
- [x] Updated Namefilter spec 

## Test plan (required)
-  Testing the Rust core.
    
    ```bash
    cargo test -p wnfs --release
    ```

- Testing the wasm bindings.
  
  ```bash
  cd crates/wasm
  ```
  
  ```bash
  yarn playwright test
  ```

## Closing issues

Fixes https://github.com/WebNativeFileSystem/rs-wnfs/issues/26
Fixes https://github.com/WebNativeFileSystem/rs-wnfs/issues/24


crates/fs/private/hamt/node.rs
```diff
@@ -169,9 +204,61 @@ where
...
+
+        // Otherwise, we go one level deep.
+        let value_index = self.get_value_index(bit_index);
+        Ok(match &self.pointers[value_index] {
+            Pointer::Values(values) => {
+                let mut node = (**self).clone();
+                let value = if values.len() == 1 {
+                    // If there is only one value, we can remove the entire pointer.
+                    node.bitmask.set(bit_index, false);
+                    let pointer = node.pointers.remove(value_index);
```
matheus23, line 230: In order to keep determinism, we have to check whether removing this pointer gets us within bucket size range. I.e. when this node contains <= 3 values, we can collapse it into a pointer and instead of linking to it, we merge it into the parent node.

crates/fs/private/namefilter/bloomfilter.rs
```diff
@@ -0,0 +1,172 @@
...
+use bitvec::prelude::BitArray;
+use xxhash_rust::xxh3;
+
+//------------------------------------------------------------------------------
+// Type Definitions
+//------------------------------------------------------------------------------
+
+/// The bloom filter is a probabilistic data structure that can be used to store a set of hashes.
+///
+/// `N` is the size of the bloom filter in bytes.
```
matheus23, line 12:
```suggestion
/// `N` is the size of the bloom filter in bytes. Must be a power of two.
```

I'm thinking we should maybe just throw that in there, since we're doing a `% (N * 8)` later on, that is supposed to be uniformly distributed.

We could *maybe* also make the parameter `N` specify the power of bits? Like `BitArray<[u8; 1 << N]>`. Dunno. Let me know what you think, this may be too many indirections until you know the actual bit size.

matheus23: Ugh, whatever, it's somewhat premature abstraction anyway, we're not using different bloom sizes. So it doesn't really matter that much.

crates/fs/private/namefilter/bloomfilter.rs
```diff
@@ -0,0 +1,172 @@
...
+    where
+        T: AsRef<[u8]>,
+    {
+        for i in self.hash_indices(item) {
+            if !self.bits[i] {
+                return false;
+            }
+        }
+
+        true
```
matheus23, line 91: I'm kind of codegolfing here, but how about this?
```suggestion
        self.hash_indices(item).all(|i| self.bits[i])
```

crates/fs/private/namefilter/namefilter.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// Adds hashes to filter until is is over the saturation threshold.
+    fn saturate(&mut self) {
+        let mut xof = {
+            let mut h = Shake256::default();
+            h.update(self.as_bytes());
+            h.finalize_xof()
+        };
+
+        while self.count_ones() <= SATURATION_THRESHOLD {
```
matheus23, line 43: Yeah, I know this is confusing, but we actually want to stay *slightly below* the saturation threshold. So once we realize that we're above the threshold, we need to "undo" the last step.

crates/fs/private/namefilter/namefilter.rs
```diff
@@ -0,0 +1,75 @@
...
+// Implementations
+//--------------------------------------------------------------------------------------------------
+
+impl Namefilter {
+    /// Creates a saturated namefilter.
+    pub fn new_saturated() -> Self {
+        let mut filter = Self::new();
+        filter.saturate();
+        filter
+    }
```
matheus23, line 33: Hm, I see how this is useful for tests, but it's not useless for any application. An empty, but saturated namefilter is essentially useless: You can't do anything with it anymore and it doesn't represent any information.

crates/fs/private/namefilter/namefilter.rs
```diff
@@ -0,0 +1,75 @@
...
+    use super::*;
+
+    #[test]
+    fn saturate_not_less_than_threshold() {
+        let namefilters = (0..100)
+            .map(|_| Namefilter::new_saturated())
+            .collect::<Vec<_>>();
+
+        for namefilter in namefilters {
+            assert!(namefilter.count_ones() >= SATURATION_THRESHOLD);
```
matheus23, line 72: Yeah, as noted above, we want the threshold to be slightly *below* the threshold. So the property is that `namefilter.saturate().count_ones()` should be `>= 990` and ` <= 1019`.

matheus23 (commented):
## HAMT

I have some comments about canonicity. It should be the case that given the same keys and values that your HAMT stores, you should always get back the same root CID, no matter what order the keys were inserted in or which intermediate operations were run. It's some quite tricky logic, but that's what gives us the really strong "if the CID is equal, the tree is equal" logic.

I think we've somewhat diverged from the original codebase now, but we can still take some inspiration. Apparently they just implement a `clean()` function, maybe that's the way to go for us, too?

https://github.com/filecoin-project/ref-fvm/blob/master/ipld/hamt/src/pointer.rs#L114

Before we put this into production we should definitely get some property tests going for this for sure, but that should be another PR :v:

## Namefilter

Saturation needs a slight change to be correct. It's described in the pseudocode in [the spec branch](https://github.com/WebNativeFileSystem/spec/blob/matheus23/initial/namefilter.md#operation-saturate).

Same thing with the property tests applies here, too. These structures have so many interesting mathematical properties, and we should write property tests to check them :) (In another PR)

crates/fs/private/namefilter/namefilter.rs
appcypher: Yeah. Shouldn't have been there.

crates/fs/private/hamt/node.rs
```diff
@@ -219,45 +222,71 @@ where
...
             Pointer::Link(link) => {
                 let child = Rc::clone(link.resolve_value(store).await?);
                 let (child, value) = child.remove_value(hashnibbles, key, store).await?;
+
                 let mut node = (**self).clone();
-                node.pointers[value_index] = Pointer::Link(Link::from(child));
+                if value.is_some() {
+                    // If something has been deleted, we attempt toc canonicalize the pointer.
+                    if let Some(pointer) =
+                        Pointer::Link(Link::from(child)).canonicalize(store).await?
```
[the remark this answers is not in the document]
appcypher: I was trying to avoid a situation where mutation happens in a call to another function and doesn't happen in the local scope. I probably took the local mutation thing too far 😅. I don't mind making it an `&mut self`.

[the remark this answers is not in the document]
appcypher: I will merge this PR but I'm going to add the change to the other PR since I have made some other changes there too.

PR #38, opened 2022-07-19
Private FS APIs

appcypher:
## Summary

This PR implements the private filesystem APIs. 

- [x] Implement core private fs APIs
  - [x] get_node
  - [x] lookup_node
  - [x] ls
  - [x] rm
  - [x] mkdir
  - [x] read
  - [x] write
- [x] Private node serialisation
- [x] Abstract Link and ReferenceableStore
- [x] Implement APIs in wasm 
- [x] Core tests and wasm browser tests

## Test plan (required)
-  Testing the Rust core.
    
    ```bash
    cargo test -p wnfs --release
    ```

- Testing the wasm bindings.
  
  ```bash
  cd crates/wasm
  ```
  
  ```bash
  yarn playwright test
  ```

## Closing issues

Fixes https://github.com/WebNativeFileSystem/rs-wnfs/issues/30
Fixes https://github.com/WebNativeFileSystem/rs-wnfs/issues/14


crates/fs/private/forest.rs
```diff
@@ -0,0 +1,176 @@
...
+
+use crate::{BlockStore, HashOutput};
+
+use super::{hamt::Hamt, namefilter::Namefilter, Key, PrivateNode, PrivateRef, NONCE_SIZE};
+
+//--------------------------------------------------------------------------------------------------
+// Type Definitions
+//--------------------------------------------------------------------------------------------------
+
+pub type EncryptedPrivateNode = (Option<Vec<u8>>, Cid); // TODO(appcypher): Change to PrivateLink<PrivateNode>.
```
matheus23, line 15: Let's also create a follow-up issue for changing this according to the spec, i.e. have the header inline, encrypted into the node and have this link just be a `Cid`.

crates/fs/private/hamt/node.rs
```diff
@@ -43,20 +46,20 @@ where
...
 where
-    K: DeserializeOwned + Serialize + AsRef<[u8]> + Clone + Eq + PartialOrd + Debug,
+    K: DeserializeOwned + Serialize + Clone + Debug + AsRef<[u8]>,
     V: DeserializeOwned + Serialize + Clone + Debug,
     H: Hasher + Clone + Debug,
 {
     /// Sets a new value at the given key.
     pub async fn set<B: BlockStore>(
-        self: Rc<Self>,
+        self: &Rc<Self>,
```
matheus23, line 55: Me learning rust: 
I'm now noticing that in some places we use `&Rc<X>`.
Shouldn't we use `Rc<X>` instead? I kind of feel like `Rc` is already like `&`, and `&Rc<X>` to me looks like doing `&&X`, which would be weird.
One problem with using `Rc<X>` is sometimes you want to have a bound on the lifetime of some return value to be defined by the lifetime of that `Rc`, but what is the `Rc`-equivalent syntax for this: `&'a X`?

crates/fs/private/key.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Encrypts the given plaintext using the key.
+    pub fn encrypt(&self, nonce_bytes: &[u8; NONCE_SIZE], data: &[u8]) -> Result<Vec<u8>> {
+        let nonce = Nonce::from_slice(nonce_bytes);
+
+        let cipher_text = Aes256Gcm::new(AesKey::from_slice(&self.0))
+            .encrypt(nonce, data)
+            .map_err(|e| FsError::UnableToEncrypt(format!("{}", e)))?;
+
+        Ok([cipher_text, nonce_bytes.to_vec()].concat())
```
matheus23, line 43: Usually you *pre*pend the nonce. That would allow an easier implementation of a streaming decryption as well.
([This stackoverflow answer](https://crypto.stackexchange.com/a/70855) explains my point much better.)

crates/fs/private/namefilter/bloomfilter.rs
```diff
@@ -106,6 +110,19 @@ impl<const N: usize, const K: usize> BloomFilter<N, K> {
...
 }
 
+impl<const N: usize, const K: usize> TryFrom<Vec<u8>> for BloomFilter<N, K> {
+    type Error = anyhow::Error;
+
+    fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
+        let bits = BitArray::<[u8; N]>::new(
+            bytes
+                .try_into()
+                .map_err(|_| anyhow!("Cannot convert vec to fixed-size array"))?,
```
matheus23, line 120: This error message could maybe say something like "Cannot convert vec to BloomFilter: Expected length {}, expected length {}"

crates/fs/private/node.rs
```diff
@@ -0,0 +1,280 @@
...
+pub enum PrivateNode {
+    File(Rc<PrivateFile>),
+    Dir(Rc<PrivateDirectory>),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PrivateRef {
+    pub(crate) saturated_name_hash: HashOutput, // Sha3-256 hash of saturated namefilter
+    pub(crate) content_key: ContentKey,         // A hash of ratchet key.
+    pub(crate) ratchet_key: RatchetKey,         // Encrypted ratchet key.
```
matheus23, line 50: I think this comment is stale now, it's not the "Encrypted" ratchet key anymore.

crates/fs/private/node.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    /// Gets the saturated namefilter for this node.
+    pub fn get_saturated_name(&self) -> Namefilter {
+        let ratchet_key = Key::new(self.ratchet.derive_key());
+        let mut name = self.bare_name.clone();
+        name.add(&ratchet_key.as_bytes());
+        name.saturate();
+        name
+    }
```
matheus23, line 100: Normally I wouldn't care about repeating some code *once*, but in this case that's a pretty crucial piece of code.
Let's extract the common part of these two functions into a third, private one.

(Also, that was me who duplicated that code when we were pairing :P Sorry!)

matheus23 (commented):
Phew. That was a big one!

Just some nitpicks & a question & some things I wanted to make sure we capture in issues for the future (I'll probably go ahead and create them right after posting this).

Let's merge this very soon(TM) :D

crates/fs/private/directory.rs
```diff
@@ -0,0 +1,773 @@
...
+        &self,
+        path_segment: &str,
+        hamt: &PrivateForest,
+        store: &B,
+    ) -> Result<Option<PrivateNode>> {
+        Ok(match self.content.entries.get(path_segment) {
+            Some(private_ref) => hamt.get(private_ref, store).await?,
+            None => None,
+        })
+    }
```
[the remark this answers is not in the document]
matheus23: https://github.com/wnfs-wg/rs-wnfs/issues/46

crates/fs/private/file.rs
```diff
@@ -0,0 +1,50 @@
...
+use super::{namefilter::Namefilter, INumber, PrivateNodeHeader};
+
+//--------------------------------------------------------------------------------------------------
+// Type Definitions
+//--------------------------------------------------------------------------------------------------
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PrivateFileContent {
+    pub(crate) metadata: Metadata,
+    pub(crate) content: Vec<u8>, // Inlined file content. // TODO(appcypher): Support linked file content.
```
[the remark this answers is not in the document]
matheus23: Actually, that's more of a TODO for me. I want to talk to Brooke about ideas for this in terms of spec before we do something in rs-wnfs right now.

crates/fs/private/forest.rs
matheus23: https://github.com/wnfs-wg/rs-wnfs/issues/47

crates/fs/private/hamt/node.rs
appcypher: Yes. It should be `Rc` unless there is a lifetime req like you mentioned

appcypher: > but what is the `Rc`-equivalent syntax for this: `&'a X`?
There is none. Rc is a runtime concept. Lifetimes are a compile-time concept. They don't work together.

crates/fs/common/utils.rs
```diff
@@ -0,0 +1,72 @@
...
+    where
+        E: serde::de::Error,
+    {
+        let bytes: [u8; N] = v.try_into().map_err(E::custom)?;
+        Ok(bytes)
+    }
+}
+
+#[cfg(test)]
+impl crate::private::Rng for TestRng {
```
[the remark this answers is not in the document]
appcypher: This looks useful. A more flexible Rng for proptests. I don't think we need `&mut self`.
The only worry I have is shrinking on seeds. Is that a useful property to test?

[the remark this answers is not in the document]
matheus23: Well, at worst it makes shrinking slow, but I agree, shrinking a seed is usually useless. However, shrinking can be turned off only some test inputs easily.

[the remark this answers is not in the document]
matheus23: > I don't think we need it to be `&mut self`.

Well, I think we do, otherwise we wouldn't be able to do what I described in the comment above with proptest's `TestRng`, because it needs `&mut self`:
https://docs.rs/proptest/latest/proptest/prelude/trait.RngCore.html#tymethod.fill_bytes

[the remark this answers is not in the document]
matheus23: To be clear: We'd need `&mut self` *for doing that "fun idea" with proptest*.
I'd say we go ahead without &mut right now!

crates/fs/common/utils.rs
```diff
@@ -43,6 +46,23 @@ impl crate::private::Rng for TestRng {
...
+        Self(proptest::test_runner::TestRng::from_seed(algorithm, seed))
+    }
+}
+
+#[cfg(test)]
+impl crate::private::Rng for ProptestRng {
+    fn random_bytes<const N: usize>(&self) -> [u8; N] {
+        use rand::RngCore;
+        let mut bytes = [0u8; N];
+        rand::thread_rng().fill_bytes(&mut bytes);
```
matheus23, line 61: Ah okay, I think there was a misunderstanding: My idea was to use `self.0.fill_bytes(&mut bytes)` here :v: (and that would require `self` to be `& mut`.
Let's just not do that for now, maybe we'll consider it for deterministic tests later.

crates/fs/private/node.rs
```diff
@@ -45,9 +45,12 @@ pub enum PrivateNode {
...
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct PrivateRef {
-    pub(crate) saturated_name_hash: HashOutput, // Sha3-256 hash of saturated namefilter
-    pub(crate) content_key: ContentKey,         // A hash of ratchet key.
-    pub(crate) ratchet_key: RatchetKey,         // Encrypted ratchet key.
+    /// Sha3-256 hash of saturated namefilter.
+    pub(crate) saturated_name_hash: HashOutput,
+    /// Content key; Sha3-256 hash of the ratchet key.
+    pub(crate) content_key: ContentKey,
+    /// Ratchet key; namefilter-derived key.
```
matheus23, line 52: Typo, I think you meant this:
```suggestion
    /// Ratchet key; skip-ratchet-derived key.
```

matheus23 (approved):

crates/fs/common/utils.rs
appcypher: Wrote it wrong. Let me fix that.

PR #39, opened 2022-07-20
Add some WASM APIs

matheus23:
- Provide metadata info in `ls`
- Expose `base_history_on` function
- Expose `load` function

Fixes https://github.com/WebNativeFileSystem/rs-wnfs/issues/34

### DIFF
--- a/crates/wasm/fs/metadata.rs
+++ b/crates/wasm/fs/metadata.rs
@@ -12,9 +12,10 @@ impl TryInto<JsValue> for JsMetadata<'_> {
     fn try_into(self) -> JsResult<JsValue> {
         let metadata = Object::new();
         let unix_meta = unix_fs_to_js_value(&self.0.unix_fs)?;
+        let version = value!(self.0.version.to_string());
 
         Reflect::set(&metadata, &value!("unixMeta"), &unix_meta)?;
-        Reflect::set(&metadata, &value!("version"), &value!("3.0.0"))?;
+        Reflect::set(&metadata, &value!("version"), &version)?;
 
         Ok(value!(metadata))
     }
--- a/crates/wasm/fs/metadata.rs
+++ b/crates/wasm/fs/metadata.rs
@@ -2,7 +2,7 @@ use crate::fs::JsResult;
 use crate::value;
 use js_sys::{Object, Reflect};
 use wasm_bindgen::JsValue;
-use wnfs::{Metadata, UnixFsMetadata, UnixFsNodeKind};
+use wnfs::{Metadata, UnixFsMetadata};
 
 pub(crate) struct JsMetadata<'a>(pub(crate) &'a Metadata);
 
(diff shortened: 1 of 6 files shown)


appcypher:
Thanks for tackling these problems. I have a few comments on some of type conversions. They are just idioms nitpicks.

crates/wasm/fs/metadata.rs
```diff
@@ -0,0 +1,44 @@
...
+
+impl TryInto<JsValue> for JsMetadata<'_> {
+    type Error = js_sys::Error;
+
+    fn try_into(self) -> JsResult<JsValue> {
+        let metadata = Object::new();
+        let unix_meta = unix_fs_to_js_value(&self.0.unix_fs)?;
+
+        Reflect::set(&metadata, &value!("unixMeta"), &unix_meta)?;
+        Reflect::set(&metadata, &value!("version"), &value!("3.0.0"))?;
```
appcypher, line 17: You missed the placeholder here.

```suggestion
        let version = value!(self.0.version.to_string());

        Reflect::set(&metadata, &value!("unixMeta"), &unix_meta)?;
        Reflect::set(&metadata, &value!("version"), &version)?;
```

crates/wasm/fs/metadata.rs
```diff
@@ -0,0 +1,44 @@
...
+fn unix_fs_kind_to_js_value(kind: &UnixFsNodeKind) -> JsValue {
+    match kind {
+        UnixFsNodeKind::Raw => value!("raw"),
+        UnixFsNodeKind::Dir => value!("dir"),
+        UnixFsNodeKind::File => value!("file"),
+        UnixFsNodeKind::Metadata => value!("metadata"),
+        UnixFsNodeKind::SymLink => value!("symlink"),
+        UnixFsNodeKind::HAMTShard => value!("hamtShard"),
+    }
+}
```
appcypher, line 44: I suggest implementing a `TryFrom<&UnixFsNodeKind>` for String type in the core impl.

crates/wasm/fs/metadata.rs
```diff
@@ -0,0 +1,44 @@
...
+
+fn unix_fs_to_js_value(unix_fs: &UnixFsMetadata) -> JsResult<JsValue> {
+    let obj = Object::new();
+    let kind = unix_fs_kind_to_js_value(&unix_fs.kind);
+
+    Reflect::set(&obj, &value!("created"), &value!(unix_fs.created))?;
+    Reflect::set(&obj, &value!("modified"), &value!(unix_fs.modified))?;
+    Reflect::set(&obj, &value!("mode"), &value!(unix_fs.mode.clone() as u32))?;
+    Reflect::set(&obj, &value!("kind"), &kind)?;
```
appcypher, line 30: If `TryFrom<UnixFsNodeKind>` is implemented for String, then it'd just be

```suggestion
    Reflect::set(&obj, &value!("kind"), &value!(String::from(&unix_fs.kind)))?;
```

matheus23:
Ok, should be good to go. Mind giving this another look @appcypher?

I don't really *like* the `sed`-based hacks in the `rs-wnfs` script, but it works. I think I want to do something that bypasses wasm-pack at some point, so we have more control over the generated `package.json`. But that'll be another PR.

appcypher (approved):
<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

<h3 align="center">Looks great! 🎉 </h3>

PR #42, opened 2022-07-28
Generate bindings with `--weak-refs` Fixes #40

### LINKED ISSUE #40: Generate JS bindings that automatically GC rust-owned structs
Currently, `PublicDirectory`s that are passed from rust to JS will have a `free()` function generated with them, which will clear up memory used inside the WASM memory & must be used to prevent memory leaks.

We should figure out a way to call `wasm-bindgen` with the [`--weak-refs` option](https://rustwasm.github.io/wasm-bindgen/reference/weak-references.html). Unfortunately `wasm-pack` [doesn't support that yet](https://github.com/rustwasm/wasm-pack/pull/937).

matheus23:
Currently will just overwrite the `.wasm` that `wasm-pack` generated by running `wasm-bindgen` and `wasm-opt` manually afterwards.

Eventually I believe we can get `--weak-refs` into `wasm-pack`.

### DIFF
--- a/scripts/rs-wnfs.sh
+++ b/scripts/rs-wnfs.sh
@@ -107,14 +107,22 @@ build_fs() {
 
 build_wasm() {
     display_header "💿 | BUILDING WASM-WNFS PROJECT | 💿"
-    echo "script_dir = $script_dir"
     cd $script_dir/../crates/wasm
     wasm-pack build --target web
-	sed -i ".bak" \
+	sed -i.bak \
         -e 's/"name": "wasm-wnfs"/"name": "wnfs",\n  "type": "module"/g' \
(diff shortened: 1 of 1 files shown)


appcypher:
@matheus23  Awesome! This fixes the GC issue. 
This runs wasm-bindgen twice but it seems fine because the rust code only compiles once.
We may want to follow your approach in the future by committing `package.json` and other files until necessary `wasm-pack` features are there.

matheus23:
Yeah, it's somewhat hacky, but it works. We can improve it later (famous last words :eyes:).

appcypher (approved):
<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

<h3 align="center">Good to go! 🎉 </h3>

PR #51, opened 2022-08-25
Fix Badges

appcypher:
Fix codecov badge


matheus23 (approved):
I think we can safely ignore CI on this one :)

Issue #54, opened 2022-08-31
Implement `mv` for the private file system


matheus23:
`mv` is much more complicated in the private file system because we need to modify the namefilters of everything in the tree that's moved.

I think it may also be useful to separate out `mv` into `cp` + `rm`.

appcypher:
In https://github.com/wnfs-wg/rs-wnfs/pull/59, I introduced an `attach` op because `cp` + `rm` for `mv` would result in more file operations. `rm` + `add` is better than a `get` + `add` + `rm`.

Now `attach` takes a node and a destination and fixes the node subtree bare names and ratchets. So `cp` = `get` + `attach`, and `mv` = `rm` + `attach`

PR #59, opened 2022-09-06
Implement `basic_mv` for Private Filesystem

appcypher:
## Summary

`mv` is much more complicated in the private file system because we need to modify the namefilters of everything in the tree that's moved.

This PR implements `mv` operation that fixes up displaced sub tree bare name ancestry and resets nodes ratchets. 

- [x] Implements `basic_mv`, `cp`, `attach` operations
- [x] Use `Rng` in constructors
- [x] Js bindings for `basic_mv` and `cp`

## Test plan (required)
-  Testing the Rust core.
    
    ```bash
    cargo test -p wnfs --release
    ```

- Testing the wasm bindings.
  
  ```bash
  cd crates/wasm
  ```
  
  ```bash
  yarn playwright test
  ```

## Closing issues

Fixes https://github.com/wnfs-wg/rs-wnfs/issues/54


crates/fs/private/node.rs
```diff
@@ -70,7 +71,77 @@ impl PrivateNode {
...
+                    working_hamt = node
+                        .update_ancestry(dir.header.bare_name.clone(), working_hamt, store, rng)
+                        .await?;
+
+                    dir.entries
+                        .insert(name.clone(), node.get_header().get_private_ref()?);
+                }
+
+                dir.header.update_bare_name(parent_bare_name);
+                dir.header.reset_ratchet(rng);
```
matheus23, line 127: Why do we reset the ratchet when we move?

I guess it makes sense for `cp`, but may not be the right thing when we're doing a `mv`. :thinking: 

Ok, I guess it's fine because if Alice `mv`s a file that Bob has access to, then (eventually we'll have a spec for that &) write a tombstone at the old position saying that we've moved the file to the new place & what ratchet it now has.

matheus23 (approved):
Great work!

crates/fs/private/node.rs
matheus23: Ah right. Great attention to the detail here :) You're right.

PR #62, opened 2022-09-23
Refactor: Remove `ReferencableStore` `Referencable` and `private/link.rs`

matheus23:
We think `Referencable*` things aren't a good abstraction between public/private, and thus it's only ever used with `Cid`s and `BlockStore`s, so we removed the abstraction.


bgins (approved):
Looks good! Thanks for the walkthrough and discussion. 🙌

PR #67, opened 2022-09-28
Make `previous` link be an array in public file system

matheus23:
Also, I've removed `get_history` from the public directory API, since now that the history may be non-linear it can't work anymore.
It's also not used anywhere & we have to figure out with what kind of API we replace it with, so I'm removing it in the mean time.


appcypher (approved):
<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

<h3 align="center">Looks good to me! 🎉 </h3>

PR #68, opened 2022-10-13
Documentation of Private Filesystem and Other Cleanups

appcypher:
## Summary

This PR adds and updates existing documentation, doc tests and makes other API clean ups.

- [x] Add doc tests
- [x] Fix API snags like Rng/RngCore diffs
- [x] Update README
- [x] Update logo

## Test plan (required)
-  Testing the Rust core.
    
    ```bash
    cargo test -p wnfs --release
    ```

- Testing the wasm bindings.
  
  ```bash
  cd crates/wasm
  ```
  
  ```bash
  yarn playwright test
  ```


README.md
```diff
@@ -31,104 +31,20 @@
 
 ##
 
-This crate is a Rust implementation of the primitives for creating and manipulating IPLD graphs that encode WNFS.
+This is a Rust implementation of [the WebNative FileSystem (WNFS) specification](https://github.com/wnfs-wg/spec). WNFS is a versioned content-addressable distributed filesystem with private and public sub systems. The private filesystem is encrypted so that only users with the right keys can access its contents. It is designed to prevent inferring metadata like the structure of the file tree. The other part of the WNFS filesystem is a simpler public filesystem that is not encrypted and can be accessed by anyone with the right address.
```
zeeshanlakhani, line 34: minor: readme wise, should we move to have links at the bottom, e.g. makes the markdown a bit easier to read https://gist.githubusercontent.com/ashmoran/1181085/raw/ba256f98b5c14cd3feb0fffab7b2638a3ca7f377/README.markdown.

README.md
```diff
@@ -31,104 +31,20 @@
...
 
-A goal of the project is to be easily compiled to WebAssembly to be used in the browsers or other environments.
+WNFS also features collaborative editing of file trees, where multiple users can edit the same tree at the same time.
+
+WNFS file trees can serialize and be deserialized from IPLD graphs with an extensible metadata section. This allows WNFS to be understood by other IPLD-based tools and systems.
```
zeeshanlakhani, line 38: minor: link IPLD graphs and or link to tools?

README.md
zeeshanlakhani: this for all readmes (since we're chatting cleanup)

crates/fs/README.md
```diff
@@ -27,118 +27,250 @@
...
+    // Print root directory.
+    println!("{:#?}", root_dir);
+}
 ```
 
-The WNFS API is immutable, therefore, we need to keep track of the updated root directory after every change.
+You may notice that we store the `root_dir` returned by the `mkdir` operation, not the `dir` we started with. That is because WNFS internal state is immutable and every operation potentially returns a new root directory. This allows us to track and rollback changes when needed. It also makes collaborative editing easier to implement and reason about. You can find more examples in the [`examples/`](examples/) folder. And there is a basic demo of the filesystem immutability [here](https://calm-thin-barista.fission.app).
 
-Each fs operation returns a possibly updated root directory that subsequent changes can be applied on.
+The private filesystem, on the other hand, is a bit more involved. [Hash Array Mapped Trie (HAMT)](https://en.wikipedia.org/wiki/Hash_array_mapped_trie) is used as the intermediate format of private file tree before it is persisted to the blockstore because HAMT helps us hide the hierarchy of the file tree.
```
zeeshanlakhani, line 193: minor: `An HAMT`

`because the HAMT`

crates/fs/examples/private.rs
```diff
@@ -0,0 +1,81 @@
+//! This example shows how to add a directory to a private forest (also HAMT) which encrypts it.
```
zeeshanlakhani, line 1: yay for examples 🥳

zeeshanlakhani (commented):
other than rebase and a few minor things, this looks great. Dope cleanup work. There's a bit more structure I want to do on pub fns w/ code comments, but that can be in another PR.

crates/fs/common/utils.rs
```diff
@@ -41,6 +42,12 @@ pub fn split_last(path_segments: &[String]) -> Result<(&[String], &String)> {
     }
 }
 
+pub fn get_random_bytes<const N: usize>(rng: &mut impl RngCore) -> [u8; N] {
```
zeeshanlakhani, line 45: we should prob comment all public comments in prep for doc generation.

zeeshanlakhani:
minor: @appcypher maybe we should make patch just `informational` vs threshold for project (related to codecov)

crates/fs/Cargo.toml
```diff
@@ -46,6 +46,7 @@ env_logger = "0.9"
 proptest = "1.0"
 test-log = "0.2"
 test-strategy = "0.2"
+rand = "0.8"
```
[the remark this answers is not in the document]
appcypher: There is no `cargo-sort` in the precommit config.

appcypher:
@zeeshanlakhani If I understand you correctly, you are suggesting removing the threshold for now?

zeeshanlakhani:
@appcypher yep, just for patch (not project)

crates/fs/Cargo.toml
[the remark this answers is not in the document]
zeeshanlakhani: @appcypher ah, right, it's in my other PR :).

zeeshanlakhani (approved):

README.md
```diff
@@ -250,6 +165,114 @@ async fn main() {
   rs-wnfs build --wasm
   ```
 
+## Usage
+
+WNFS does not have an opinion on where you want to persist your content or the file tree. Instead, the API expects any object that implements the async [`BlockStore`][blockstore-trait] interface. This implementation also defers system-level operations to the user; requiring that operations like time and random number generation be passed in from the interface. This makes for a clean wasm interface that works everywhere.
+
+Let's see an example of working with a public directory. Here we are going to use the memory-based blockstore provided by library.
```
matheus23, line 172:
```suggestion
Let's see an example of working with a public directory. Here we are going to use the memory-based blockstore provided by the library.
```

README.md
```diff
@@ -250,6 +165,114 @@ async fn main() {
...
+    let PrivateOpResult { result, .. } = root_dir
+        .ls(&["pictures".into()], true, hamt, store)
+        .await
+        .unwrap();
+
+    println!("Files in /pictures: {:#?}", result);
+}
+```
+
+Namefilters are currently how we represent the identity key of a node in the filesystem. They have nice properties, one of which is the ability to check if one node belongs to another. This is necessary in a filesystem where metadata like hierarchy needs to be hidden from observing agents. One notable caveat with namefilters is that they can only reliably store information of a file tree 47 levels deep or less so there is a plan to replace them with cryptographic accumlators in the near future.
```
matheus23, line 272:
```suggestion
Namefilters are currently how we represent the identity key of a node in the filesystem. They have nice properties, one of which is the ability to check if one node belongs to another. This is necessary in a filesystem where metadata like hierarchy needs to be hidden from observing agents. One notable caveat with namefilters is that they can only reliably store information of a file tree 47 levels deep or less so there is a plan to replace them with other cryptographic accumlators in the near future.
```

Technically namefilters are already cryptographic accumulators. But, really this change is a technicality.

crates/fs/README.md
```diff
@@ -27,118 +27,129 @@
...
-## Building the Project
-
-- Build project
-
-  ```bash
-  cargo build --release
-  ```
-
-## Testing the Project
+Namefilters are currently how we represent the identity key of a node in the filesystem. They have nice properties, one of which is the ability to check if one node belongs to another. This is necessary in a filesystem where metadata like hierarchy needs to be hidden from observing agents. One notable caveat with namefilters is that they can only reliably store information of a file tree 47 levels deep or less so there is a plan to replace them with cryptographic accumlators in the near future.
```
matheus23, line 144:
```suggestion
Namefilters are currently how we identify private node blocks in the filesystem. They have nice properties, one of which is the ability to check if one node belongs to another. This is necessary in a filesystem where metadata like hierarchy needs to be hidden from observing agents. One notable caveat with namefilters is that they can only reliably store information of a file tree 47 levels deep or less so there is a plan to replace them with other cryptographic accumlators in the near future.
```

I'd use "node blocks" instead of "node", because each namefilter only identifies a single block. The next revision will have a different identifier.
The identifier for nodes themselves would be `inumbers`. 

(The same paragraph may exist somewhere else and could maybe use the same suggestion?)

crates/fs/common/error.rs
```diff
@@ -56,8 +56,11 @@ pub enum FsError {
     #[error("Expected bare ratchet key")]
     ExpectBareRatchetKey,
 
-    #[error("Missing unix fs metadata field")]
-    MissingUnixFSMetadata,
+    #[error("Missing created time metadata")]
+    MissingCreatedTimeMetadata,
+
+    #[error("Missing modified time metadata")]
+    MissingModifiedTimeMetadata,
```
matheus23, line 63: No change for this PR needed - but we should probably make timestamps in the metadata optional.

crates/fs/common/metadata.rs
```diff
@@ -31,7 +33,7 @@ pub struct Metadata(pub BTreeMap<String, Ipld>);
 impl Metadata {
     /// Creates a new metadata representing a UnixFS node.
     pub fn new(time: DateTime<Utc>) -> Self {
-        let time = time.timestamp();
+        let time = time.timestamp_nanos();
```
matheus23, line 36: Why `timestamp_nanos` now?
I'd rather have them be POSIX-compatible second-based timestamps, which is probably what most people familiar with POSIX interfaces would expect (especially since we don't have anything indicating the unit here).
It's especially important for JS, where 64-bit integers aren't a thing.

crates/fs/private/forest.rs
```diff
@@ -3,17 +3,34 @@ use std::rc::Rc;
...
-use super::{hamt::Hamt, namefilter::Namefilter, Key, PrivateNode, PrivateRef, Rng};
+use super::{hamt::Hamt, namefilter::Namefilter, Key, PrivateNode, PrivateRef};
 
 //--------------------------------------------------------------------------------------------------
 // Type Definitions
 //--------------------------------------------------------------------------------------------------
 
-// TODO(appcypher): Change Cid to PrivateLink<PrivateNode>.
-// TODO(appcypher): And eventually to BTreeSet<PrivateLink<PrivateNode>>.
+/// PrivateForest is a HAMT that stores CIDs of private nodes using their saturated namefilters as keys.
```
matheus23, line 16:
```suggestion
/// PrivateForest is a HAMT that stores CIDs of encrypted private nodes keyed by saturated namefilters.
```

I want to add the encryption thing in there since I think that may help in the overview. And needed to change the wording around "key" to make it obvious it's not encrypted using saturated namefilters, but keyed by them.

crates/fs/private/forest.rs
```diff
@@ -3,17 +3,34 @@ use std::rc::Rc;
...
+///
+/// On insert, nodes are serialized to DAG CBOR and encrypted with their private refs and then stored in
+/// a accompanying blockstore. And on lookup, the nodes are decrypted and deserialized with the same private
```
matheus23, line 19:
```suggestion
/// an accompanying blockstore. And on lookup, the nodes are decrypted and deserialized with the same private
```

crates/fs/private/hamt/hash.rs
```diff
@@ -16,11 +16,38 @@ const MAX_CURSOR_DEPTH: usize = HASH_BYTE_SIZE * 2;
...
+/// # Examples
+///
+/// ```
+/// use wnfs::private::HashNibbles;
+///
+/// let nibbles = HashNibbles::new(&[0xFFu8; 32]);
+/// println!("{:?}", nibbles);
+/// ```
 #[derive(Debug, Clone)]
 pub struct HashNibbles<'a> {
```
matheus23, line 52: Does this actually need to be a `pub` thing?
I guess now it's documented and we can keep it at that, but just wondering whether we want to export the guts of the HAMT.

matheus23 (commented):
Some small notes, wow, so many docs :eyes: :books: :sparkles:

crates/fs/common/metadata.rs
appcypher: Wasn't aware of this constraint. Let me fix that.
Was having troubles converting `time.timestamp()` back to `Utc`. Chronos seems to represent time to its nanoseconds, and that was  breaking the equality checks.

crates/fs/common/error.rs
appcypher: Right. It is an easy change to make. Let me fix it.

matheus23 (approved):

Issue #17, opened 2022-05-09
Add Benchmark Tests


appcypher:
It is useful to have a benchmark comparison of ts-based wnfs implementation and wasm-wnfs.

It will help us have an idea of the performance gap.  And also identify any improvement that can be made.

The test will is also important for detecting regression.

PR #75, opened 2022-10-21
Initial Benchmark Work

appcypher:
## Summary

This PR adds benchmarks for `rs-wnfs` namefilter and hamt implementations as an intial step towards having benchmarks for important features of WNFS. There is also the [client-side benchmarking][client-bench] that is not covered here.

## Test plan (required)
-  Testing the Rust core.
    
    ```bash
    cargo test -p wnfs --release
    ```

-  Run benchmarks

    To run all benchmarks:
    
    ```bash
    cargo bench -p wnfs-bench
    ```
    
    And if you have gnuplot installed:
    
    ```
    cargo bench -q -p wnfs-bench -- --plotting-backend gnuplot
    ```

[client-bench]: https://github.com/wnfs-wg/rs-wnfs/issues/17


crates/fs/Cargo.toml
```diff
@@ -46,6 +46,7 @@ env_logger = "0.9"
 proptest = "1.0"
 test-log = "0.2"
 test-strategy = "0.2"
+criterion = { version = "0.4.0", features = ["async_std"] }
```
zeeshanlakhani, line 49: minor: just 0.4, let the patch move.

crates/fs/benches/hamt.rs
```diff
@@ -0,0 +1,164 @@
...
+        )?
+        black_box(Hamt::with_root(root))
+    }};
+}
+
+fn hamt_load_decode(c: &mut Criterion) {
+    let mut store = MemoryBlockStore::default();
+    let cid: Cid = async_std::task::block_on(async {
+        let hamt: Hamt<String, _> = hamt!(&mut store, {
+            ("foo".into(), 1),
```
zeeshanlakhani, line 103: any magic number of 13? should we decode hamts at different sizes? We could maybe use proptest to generate various numbers (did something in previous libs).

crates/fs/benches/hamt.rs
```diff
@@ -0,0 +1,164 @@
...
+            let root = root.set($key, $value, $store).await.unwrap();
+            $(
+                let root = root.set($key_rest, $value_rest, $store).await.unwrap();
+            )*
+        )?
+        black_box(Hamt::with_root(root))
+    }};
+}
+
+fn hamt_load_decode(c: &mut Criterion) {
```
zeeshanlakhani, line 99: also do we want to measure throughput on these kinda tests? e.g. https://bheisler.github.io/criterion.rs/book/user_guide/advanced_configuration.html

crates/fs/Cargo.toml
zeeshanlakhani, line 49: more of interesting note: https://www.tweag.io/blog/2022-03-03-criterion-rs/ (criterion can run wasm benchmarks, if you turn default_features off cause rayon)

crates/fs/benches/namefilter.rs
```diff
@@ -0,0 +1,72 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
```
zeeshanlakhani, line 1: we should prob measure and compare in CI. One example action out there is https://github.com/boa-dev/criterion-compare-action, but we can build it ourselves using critcmp: https://github.com/BurntSushi/critcmp. Actually, many use https://github.com/benchmark-action/github-action-benchmark, so maybe that's the way to go these days. 

As an example of the github-action-benchmark and also uploading of criterion artifacts (which are nice): https://github.com/osohq/oso/blob/main/.github/workflows/bench.yml#L29.

crates/fs/benches/hamt.rs
appcypher: No reason at all. I should probably remove this static version in favor of a loop.
What do you think?
Also have you used `proptest` with criterion before? How do you deal with non-determinism?

zeeshanlakhani: so, used a deterministic rvg, with a seed across runs: https://github.com/capsule-rs/capsule/blob/410696acb2e033cabb287750b36810753b54a59e/core/src/testils/rvg.rs#L38. We also created our own criterion iters for allowing for generated batches: https://github.com/capsule-rs/capsule/blob/410696acb2e033cabb287750b36810753b54a59e/core/src/testils/criterion.rs#L98.

zeeshanlakhani: @appcypher a loop is prob fine; I was wondering if we wanted to decode across a set of different hamt sizes?

crates/fs/Cargo.toml
appcypher: That's great to know but I think it's fine to ignore criterion for wasm, just like we do for the tests and logging. Right now we only test our wasm blob from the Js side and I expect benchmarking will also happen the same way. I think that is a more relevant way of testing everything works from a user perspective. Unless someone has a non-web use case for wnfs, then it is easy to expose what they need behind feature flags.

appcypher: @zeeshanlakhani Also. Curious. What is benefit of not providing the patch?

zeeshanlakhani: @appcypher typically libs shouldn't break compat at the patch level, so when building you'll automatically get the newest minor in your lock file.

zeeshanlakhani: There are obvs bad deps/libs out there tho :).

wnfs/README.md
```diff
@@ -1,6 +1,6 @@
 <div align="center">
   <a href="https://github.com/wnfs-wg" target="_blank">
-    <img src="../assets/logo.png" alt="Fission Logo" width="100" height="100"></img>
+    <img src="https://raw.githubusercontent.com/wnfs-wg/rs-wnfs/main/assets/logo.png" alt="WNFS Logo" width="100" height="100"></img>
```
zeeshanlakhani, line 3: @appcypher any reason not to go relative?

appcypher: Because it breaks on [crates.io](https://crates.io/crates/wnfs) since the parent workspace isn't published.

zeeshanlakhani: @appcypher derp of course. my bad.

wnfs/README.md
```diff
@@ -149,7 +149,6 @@ Check the [`examples/`][wnfs-examples] folder for more examples.
 [hamt-wiki]: https://en.wikipedia.org/wiki/Hash_array_mapped_trie
 [ipld-spec]: https://ipld.io/
 [npm-ipld-tools]: https://www.npmjs.com/search?q=ipld
-[wnfs-examples]: examples/
+[wnfs-examples]: https://github.com/wnfs-wg/rs-wnfs/tree/main/wnfs/examples
```
zeeshanlakhani, line 152: 👍🏽

wnfs/benches/namefilter.rs
```diff
@@ -0,0 +1,74 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion, Throughput};
+use wnfs::{dagcbor, private::namefilter::Namefilter};
+
+fn add(c: &mut Criterion) {
+    c.bench_function("namefilter add", |b| {
+        b.iter(|| {
+            let mut namefilter = black_box(Namefilter::default());
```
matheus23, line 7: Same thing here. We could/should split the initialization out with `iter_batched`.

wnfs/benches/namefilter.rs
```diff
@@ -0,0 +1,74 @@
...
+
+fn saturate(c: &mut Criterion) {
+    let mut namefilter = Namefilter::default();
+    for i in 0..50 {
+        namefilter.add(&i.to_string());
+    }
+
+    c.bench_function("namefilter saturate", |b| {
+        b.iter(|| {
+            black_box(namefilter.saturate());
```
matheus23, line 38: I'm not sure whether this is actually doing anything for two reasons:
- I worry that `namefilter.saturate()` is detected to just be a constant value and thus optimized. (Do we want to put a `black_box(namefilter).saturate()` here instead?)
- Since that's a 2048-bit bloom filter with 30 hashes & we're filling it over the half-way line with 50 elements, it actually is already over-saturated and saturation will just detect that and bail immediately. This namefilter should be filled with some random element and we should test the worst case performance of saturating an almost empty namefilter.

matheus23 (commented):
I worry there's some `black_box` issues if we don't spam these furiously around our codebase, *or* if we don't randomly generate the inputs at runtime.

wnfs/benches/namefilter.rs
appcypher: Oops. Totally missed that.

wnfs/benches/hamt.rs
```diff
@@ -0,0 +1,149 @@
...
+            .await
+            .unwrap();
+
+        let cid = store.put_serializable(&encoded_hamt).await.unwrap();
+
+        (cid, encoded_hamt)
+    });
+
+    let mut group = c.benchmark_group("node load and get");
+    group.throughput(Throughput::Bytes(bytes.len() as u64));
```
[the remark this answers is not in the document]
zeeshanlakhani: @matheus23 yeah, I was also mentioning various sizes of HAMTs (some interval).

zeeshanlakhani:
FYI: We should link to https://wnfs-wg.github.io/rs-wnfs/dev/bench/ somewhere in the README (where the benchmarks are output per commit).

.github/workflows/bench.yaml
```diff
@@ -0,0 +1,41 @@
...
+        run: cargo bench -q -p wnfs -- --output-format bencher | tee output.txt
+
+      - name: Store Benchmark Result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: Rust Benchmark
+          tool: 'cargo'
+          output-file-path: output.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
```
zeeshanlakhani, line 37: @appcypher maybe we shouldn't auto push directly: https://github.com/benchmark-action/github-action-benchmark#run-only-on-your-branches, but do something like

```
auto-push: ${{ github.event_name == 'push' && github.repository == 'wnfs-wg/rs-wnfs' && github.ref == 'refs/heads/main' }}
```

.github/workflows/bench.yaml
```diff
@@ -0,0 +1,41 @@
...
+          override: true
+          toolchain: stable
+          profile: minimal
+
+      - name: Cache Project
+        uses: Swatinem/rust-cache@v2
+
+      - name: Run benchmark
+        run: cargo bench -q -p wnfs -- --output-format bencher | tee output.txt
+
```
zeeshanlakhani, line 29: Also, should prob artifact the data anyway (esp for PRs):

```
name: Upload benchmark result artifact
        uses: actions/upload-artifact@v2
        with:
          name: bench_result
          path: output.txt
```

wnfs/benches/hamt.rs
```diff
@@ -0,0 +1,149 @@
...
+fn node_set(c: &mut Criterion) {
+    c.bench_function("node set", |b| {
+        b.to_async(AsyncStdExecutor).iter(|| async {
+            let store = &mut MemoryBlockStore::default();
+            let mut node = black_box(Rc::new(<Node<_, _>>::default()));
+
+            for i in 0..50 {
+                node = black_box(node.set(i.to_string(), i, store).await.unwrap());
+            }
+        })
```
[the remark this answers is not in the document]
appcypher: Unfortunately, the async version of `iter_batched` does not support async setup fns, but I can still get it to work with `async_std::task::block_on`.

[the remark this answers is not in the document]
appcypher: Also the mutable store makes things more complicated. Not sure I can refactor this in a clean way without cloning store.

wnfs/benches/namefilter.rs
```diff
@@ -1,62 +1,94 @@
...
+
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion, Throughput};
+use proptest::{arbitrary::any, collection::vec, test_runner::TestRunner};
+use sampleable::Sampleable;
 use wnfs::{dagcbor, private::namefilter::Namefilter};
 
+const FILTER_CAPACITY: usize = 47;
+
 fn add(c: &mut Criterion) {
+    let mut runner = TestRunner::deterministic();
```
zeeshanlakhani, line 11: +1

wnfs-bench/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
```
zeeshanlakhani, line 1: 👍🏽

wnfs/private/hamt/strategies.rs
```diff
@@ -0,0 +1,165 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use std::{collections::HashMap, fmt::Debug, hash::Hash, rc::Rc};
+
+#[derive(Debug, Clone)]
+pub enum Operation<K, V> {
+    Insert(K, V),
+    Remove(K),
+}
+
+impl<K, V> Operation<K, V> {
```
zeeshanlakhani, line 13: minor: even though gated for tests, it's on by feature, so we should doc these pubs.

matheus23 (approved):
Some minor things but otherwise good to go :)

zeeshanlakhani:
> FYI: We should link to https://wnfs-wg.github.io/rs-wnfs/dev/bench/ somewhere in the README (where the benchmarks are output per commit).

didn't see this link (know it only runs on main now).

zeeshanlakhani (approved):
minor the link comment

Issue #74, opened 2022-10-19
Private Multivalues


matheus23:
Essentially have the PrivateForest's values be an array of CIDs.

PR #78, opened 2022-10-26
Make private forest multivalue

matheus23:
This PR only changes the data format, but minimally changes the actual data stored. Specifically: We never actually generate multivalues in rs-wnfs. So why add support for them?
The reason is being future-proof: We *know* we'll make use of multivalues in the future, so we add support for them *today*, so that *today*'s version of rs-wnfs will have *some* support for them even when future versions of rs-wnfs actually write multivalues.

In this PR:
- the `PrivateForest` type alias changed to use `BTreeSet<Cid>` as value, instead of just `Cid`
- `PrivateForest.set` was renamed to `put`, and it adds to the multivalues at a given key, instead of replacing. There's a performance optimization opportunity here to avoid walking the HAMT twice by creating a `modify`/`upsert` implementation in `hamt/node.rs`. But that turned out to be quite some effort, so I'm postponing that for now.
- `PrivateForest.get` now expects a `resolve_bias` function parameter. Today that mostly ends up being `BTreeSet::first`, in order to fetch the smallest CID. That's the tie-breaking that's specified for WNFS by default, but that will be something different e.g. when we have previous pointers, then we'll bias towards those pointers.
- `PrivateForest.get_encrypted` returns the whole `BTreeSet` now.
- There's some helper functions in `PrivateForest` like `resolve_lowest`, `resolve_single` and `resolve_one_of` that are supposed to be helpers for use with `PrivateForest.get`.

Closes issue: #74

### DIFF
--- a/wnfs/examples/private.rs
+++ b/wnfs/examples/private.rs
@@ -33,7 +33,7 @@ async fn main() {
 
     // Fetch and decrypt a directory from the private forest using provided private ref.
     let dir = forest
-        .get(&private_ref, BTreeSet::first, store)
+        .get(&private_ref, PrivateForest::resolve_lowest, store)
         .await
         .unwrap();
 
--- a/wnfs/examples/private.rs
+++ b/wnfs/examples/private.rs
@@ -4,7 +4,7 @@
 use chrono::Utc;
 use libipld::Cid;
 use rand::{thread_rng, RngCore};
-use std::rc::Rc;
+use std::{collections::BTreeSet, rc::Rc};
 use wnfs::{
     dagcbor,
     private::{PrivateForest, PrivateRef},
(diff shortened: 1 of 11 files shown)


wnfs/examples/private.rs
```diff
@@ -4,7 +4,7 @@
 use chrono::Utc;
 use libipld::Cid;
 use rand::{thread_rng, RngCore};
-use std::rc::Rc;
+use std::{collections::BTreeSet, rc::Rc};
```
appcypher, line 7: Unused import

appcypher: Same for other doc examples

appcypher (approved):
Looks good to me! 🎉 
Only minor fixes to be addressed.

wnfs/examples/private.rs
appcypher: No clue. Looks unresolved for now. https://github.com/rust-lang/cargo/issues/6424

appcypher: cc @zeeshanlakhani. Any idea?

PR #82, opened 2022-11-02
use src for lib.rs rust files

zeeshanlakhani:
Just a directory fix to follow what's normal for lib.rs and rust files: https://doc.rust-lang.org/cargo/guide/project-layout.html. Also, remove `crate-type` for rust-native code.


matheus23 (approved):
Alright :+1:

dignifiedquire:
💙

PR #85, opened 2022-11-07
Use `Rc::try_unwrap` for (possibly?) better performance

matheus23:
Putting this up here for discussion.

We can make use of mutability in `Node::set_value` and `Node::remove_value`, by turning values of type `Rc<Node<...>>` into `mut Node<...>` using `Rc::try_unwrap(...).unwrap_or_else(|rc| (*rc).clone())`.
Essentially that'll just give us an owned `Node` if we hold the only reference to it, or it'll clone the inner `Rc`.
Theory is that if the calling code is basically doing lots of changes on the HAMT in a tight loop, that's going to get sped up, since that'll end up being lots of in-place mutation.

The benchmarks don't look *too* promising though. I got ~10% improvements.
To test this, just compare the two commits in the PR against each other. Specifically their `cargo bench -p wnfs-bench node`.

Also, I had to hand-expand the `async_recursion` macro for `set_value` and `remove_value`, because otherwise there would've been lifetime issues with the original versions. We should probably figure out another way, but only if we really want to keep this PR.

### DIFF
--- a/wnfs/src/private/hamt/node.rs
+++ b/wnfs/src/private/hamt/node.rs
@@ -399,34 +399,33 @@ where
             let mut node = Rc::try_unwrap(self).unwrap_or_else(|rc| (*rc).clone());
 
             let removed = match &mut node.pointers[value_index] {
-                Pointer::Values(values) => {
-                    if values.len() == 1 {
-                        // If the key doesn't match, return without removing.
-                        if &H::hash(&values[0].key) != hashnibbles.digest {
-                            None
-                        } else {
-                            // If there is only one value, we can remove the entire pointer.
-                            node.bitmask.set(bit_index, false);
-                            match node.pointers.remove(value_index) {
-                                Pointer::Values(mut values) => Some(values.pop().unwrap()),
-                                _ => unreachable!(),
-                            }
-                        }
+                // If there is only one value, we can remove the entire pointer.
+                Pointer::Values(values) if values.len() == 1 => {
+                    // If the key doesn't match, return without removing.
+                    if &H::hash(&values[0].key) != hashnibbles.digest {
+                        None
                     } else {
-                        // Otherwise, remove just the value.
-                        match values
-                            .iter()
-                            .position(|p| &H::hash(&p.key) == hashnibbles.digest)
-                        {
-                            Some(i) => {
-                                let value = values.remove(i);
-                                // We can take here because we replace the node.pointers here afterwards anyway
-                                let values = std::mem::take(values);
-                                node.pointers[value_index] = Pointer::Values(values);
-                                Some(value)
-                            }
-                            None => None,
+                        node.bitmask.set(bit_index, false);
+                        match node.pointers.remove(value_index) {
+                            Pointer::Values(mut values) => Some(values.pop().unwrap()),
+                            _ => unreachable!(),
                         }
+                    }
+                }
+                // Otherwise, remove just the value.
+                Pointer::Values(values) => {
+                    match values
+                        .iter()
+                        .position(|p| &H::hash(&p.key) == hashnibbles.digest)
+                    {
+                        Some(i) => {
+                            let value = values.remove(i);
+                            // We can take here because we replace the node.pointers here afterwards anyway
+                            let values = std::mem::take(values);
+                            node.pointers[value_index] = Pointer::Values(values);
+                            Some(value)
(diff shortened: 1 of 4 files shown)


matheus23:
@appcypher and I figured out that adding `+ 'static` to `K`, `V` and `H` type parameters would make the lifetime annotations redundant.

It's for sure safe to annotate `H: 'static`, but for `K` and `V` it's not *too* obvious. It'd limit users of `Node<K, V, H>` to only put owned things (never references e.g. no `&str` for keys) into the HAMT. That's the case for all of rs-wnfs, so technically we could just do that? That'd make the code cleaner again and we could use `#[async_recursion(?Send)]` again.

appcypher (approved):
<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

And like you suggested, maybe we add `'static` for H at the impl level. Doesn't matter much either way.

wnfs/src/private/hamt/node.rs
```diff
@@ -376,76 +377,81 @@ where
...
-                    {
-                        node.pointers[value_index] = pointer;
+
+            let value_index = self.get_value_index(bit_index);
+
+            let mut node = Rc::try_unwrap(self).unwrap_or_else(|rc| (*rc).clone());
+
+            let removed = match &mut node.pointers[value_index] {
+                Pointer::Values(values) => {
+                    if values.len() == 1 {
```
zeeshanlakhani, line 404: this nest of nested matches and conditionals is a bit gnarly? Anyway, to break these up?

wnfs/src/private/hamt/node.rs
```diff
@@ -376,76 +377,81 @@ where
...
-                        .map(|i| {
-                            let value = values.remove(i);
-                            node.pointers[value_index] = Pointer::Values(values);
-                            value
-                        })
-                };
-
-                (Rc::new(node), value)
+            // If the bit is not set yet, return None.
+            if !self.bitmask[bit_index] {
```
zeeshanlakhani, line 394: is this typically the bitarray under the hood? can you not use get, which exposes an option and use some option fns magic to return the result w/ a `?`. Always prefer the `?` over the early return.

wnfs/src/private/hamt/node.rs
```diff
@@ -376,76 +377,81 @@ where
...
-    async fn remove_value<'a, 'b, B: BlockStore>(
-        self: &'a Rc<Self>,
-        hashnibbles: &'b mut HashNibbles,
-        store: &B,
-    ) -> Result<(Rc<Self>, Option<Pair<K, V>>)>
+    fn remove_value<'a, B: BlockStore>(
+        self: Rc<Self>,
+        hashnibbles: &'a mut HashNibbles,
+        store: &'a B,
+    ) -> LocalBoxFuture<'a, Result<RemovedResult<K, V, H>>>
```
zeeshanlakhani, line 384: tbh, I'm almost more preferring this over the macro, which hides a bit too much.

matheus23: Hm. Alright. Didn't expect that :)

I'll remove the `RemovedResult` type alias and just `#[allow(clippy...)]` though.

wnfs/src/private/hamt/node.rs
matheus23: Actually, nevermind. There's lots of array bittwiddling happening in the HAMT implementation.
Any invalid access should be a show-stopping bug in the implementation.
I coded up something, but really didn't like it. Instead, I've added some tests to the deserialization process that makes sure you never decode something from the outside that has an invalid format. Thus, this should error out at the right point in time now.
I'm fairly confident the property based tests make sure we never actually break the invariants from within this implementation (i.e. we can skip bounds checks after we've validated the invariants from deserialized nodes).

wnfs/src/private/hamt/node.rs
matheus23: Hm. Hard to do. The most straightforward idea would be to extract all this functionality into the `Pointer` type, but then we'd have mutually recursive async functions, which doesn't really work.

Instead I've opted to merge one level of the `if` into the match, I like that version a little better at least.

zeeshanlakhani: me too.

zeeshanlakhani (approved):

Issue #80, opened 2022-10-31
Private File Sharding 


walkah:

b5:
wat

matheus23:
Private file sharding as described in https://github.com/wnfs-wg/spec/pull/44

PR #88, opened 2022-11-10
Private File Sharding

appcypher:
## Summary

This PR implements sharding of file content; breaking the content up and storing them in the HAMT if they are larger than certain threshold, storing them inline otherwise.

- [x] Private file sharding
- [x] Stream shards
- [x] Wasm APIs
- [x] Tests

## Test plan (required)
-  Testing
    
    ```bash
    rs-wnfs test --all
    ```

## Closing issues

Fixes https://github.com/wnfs-wg/rs-wnfs/issues/80


appcypher:
Just made our tests slower; enabled Safari and Edge for the wasm-js-tests. We've been ignoring them for a while. I'm considering disabling Edge since we have Chrome tests.

appcypher:
Made it faster switching to `macos-latest`. Still not as fast as I'd want so removing Edge tests is still on the table.

wnfs/src/private/file.rs
```diff
@@ -62,29 +108,370 @@ impl PrivateFile {
...
+    ///     )
+    ///     .await
+    ///     .unwrap();
+    ///
+    ///     let mut content_at_0 = file.get_block_at(0, &hamt, store).await.unwrap();
+    ///
+    ///     assert_eq!(content, content_at_0.unwrap());
+    /// }
+    /// ```
+    pub async fn get_block_at<B: BlockStore>(
```
matheus23, line 343: What do you think of a `seek` API here?
I.e. `seek(&self, offset: usize, limit: usize) -> Option<Vec<u8>>`, roughly?
It would return `None` if `offset >= file_size` and have the invariant that `vec.len() <= limit`, if the result is `Some(vec)`.

I guess we could have both, and `seek` could be implemented in terms of `get_block_at`?
Or we could also have `stream_from` take a block index as offset, so we would stream blocks starting at that block index, which could be the ultimate underlying function?

Thoughts?

wnfs/src/private/file.rs
```diff
@@ -62,29 +108,370 @@ impl PrivateFile {
...
+
+    /// Determines where to put the content of a file. This can either be inline or stored up in chunks in a HAMT.
+    pub(super) async fn prepare_content<B: BlockStore, R: RngCore>(
+        bare_name: &Namefilter,
+        content: Vec<u8>,
+        mut hamt: Rc<PrivateForest>,
+        store: &mut B,
+        rng: &mut R,
+    ) -> Result<(FileContent, Rc<PrivateForest>)> {
+        // TODO(appcypher): Use a better heuristic to determine when to use external storage.
```
matheus23, line 379: Still a remaining TODO?

Perhaps its just safer to always go external content by default?

I'm fine with leaving this here, just wanted to raise this in case it was accidental. I think it's unlikely that metadata on a node exceeds ~50kb.

wnfs/src/private/file.rs
```diff
@@ -62,29 +108,370 @@ impl PrivateFile {
...
+        iter::from_fn(move || {
+            if block_count < 1 {
+                return None;
+            }
+
+            let label = Self::create_block_label(key, block_count, bare_name);
+            block_count -= 1;
+            Some(label)
+        })
+    }
```
matheus23, line 461: If I read this correctly, this generates the block labels in reverse, i.e. starting from `block_count` then `block_count-1` and ending with `0`.

This would mean that the first block of bytes for a file would be found at the accumulator of `parent_accumulator + key + hash(key ++ block_count)` instead of `parent_accumulator + key + hash(key ++ 0)`.

If that's the case, then this should change to match the spec.

```suggestion
    /// Generates the labels for the shards of a file.
    fn generate_shard_labels<'a>(
        key: &'a Key,
        block_count: usize,
        bare_name: &'a Namefilter,
    ) -> impl Iterator<Item = Namefilter> + 'a {
        let mut index = 0;
        iter::from_fn(move || {
            if index >= block_count {
                return None;
            }

            let label = Self::create_block_label(key, index, bare_name);
            index += 1;
            Some(label)
        })
    }
```

matheus23 (changes requested):
Only small things except for the shard order thing :)

wnfs/src/private/file.rs
appcypher: You are right. Thought of it as an optimization but it breaks the spec. 
Will fix.

wnfs/src/private/file.rs
appcypher: What would be the benefit of a limit in bytes?

matheus23 (approved):
Letsgooooo
One step closer to alpha!

Excited birthday girl

wnfs-wasm/src/fs/utils.rs
```diff
@@ -67,11 +77,20 @@ pub(crate) fn create_private_op_result<T: Into<JsValue>>(
...
+    hamt: Rc<WnfsPrivateForest>,
+) -> JsResult<JsValue> {
+    let op_result = Object::new();
+
+    Reflect::set(
+        &op_result,
+        &value!("file"),
+        &PrivateFile(Rc::new(file)).into(),
+    )?;
+    Reflect::set(&op_result, &value!("hamt"), &PrivateForest(hamt).into())?;
```
zeeshanlakhani, line 91: dumb q: why not return the hamt directly, instead of keyed to "hamt" in an JS object? Feels strange to label the "hamt" like this, but I'm probably missing something.

wnfs-wasm/src/fs/private/file.rs
```diff
@@ -1,42 +1,87 @@
...
 // Implementations
 //--------------------------------------------------------------------------------------------------
 
 #[wasm_bindgen]
 impl PrivateFile {
-    /// Creates a new private file.
-    #[wasm_bindgen(constructor)]
-    pub fn new(
+    /// Creates an empty private file.
+    pub fn empty(parent_bare_name: Namefilter, time: &Date, mut rng: Rng) -> JsResult<PrivateFile> {
```
zeeshanlakhani, line 32: minor: this still feels like new to me, even though it's essentially empty to start, b/c in Rust-land new can also mean an empty hashmap etc, but just thinking aloud.

wnfs-wasm/src/fs/private/file.rs
```diff
@@ -1,42 +1,87 @@
...
-use crate::fs::{JsResult, Namefilter, Rng};
-
 //--------------------------------------------------------------------------------------------------
 // Type Definitions
 //--------------------------------------------------------------------------------------------------
 
 /// A file in a WNFS public file system.
 #[wasm_bindgen]
-pub struct PrivateFile(WnfsPrivateFile);
+pub struct PrivateFile(pub(crate) Rc<WnfsPrivateFile>);
```
zeeshanlakhani, line 23: maybe I'm missing it in the examples, but does this really need to be pub(crate) inside? Maybe point me to where it's used as pub?

wnfs-wasm/src/fs/private/file.rs
```diff
@@ -1,42 +1,87 @@
...
+        Ok(future_to_promise(async move {
+            let (file, hamt) = WnfsPrivateFile::with_content(
+                parent_bare_name.0,
+                time,
+                content,
+                hamt.0,
+                &mut store,
+                &mut rng,
+            )
+            .await
```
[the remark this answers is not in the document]
matheus23: `with_content` returns a `Promise`, and `future_to_promise` turns an `impl Future` into a `Promise`.

It's similar to how you can write non-async functions if you just return an `impl Future` "manually".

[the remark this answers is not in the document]
zeeshanlakhani: gotcha. Yeah, I knew what future_to_promose did, but see what you mean.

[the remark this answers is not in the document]
appcypher: `future_to_promise` is the magic. It wraps the future in a closure passed to a promise. When the promise is resolved or rejected the future is spawned and run a task scheduler.

wnfs-wasm/src/fs/private/file.rs
appcypher: In https://github.com/wnfs-wg/rs-wnfs/blob/appcypher/private-file-sharding/wnfs-wasm/src/fs/utils.rs#L81. But since it is only really used in one place I should probably remove the pub(crate) and pass in the PrivateFile itself.

wnfs-wasm/src/fs/utils.rs
appcypher: BTW, we can't simply return an hamt. We need to somehow convert it to a JsValue and Reflect.set is the only way I know of to do that with wasm-bindgen. On that note, I can change it to an Array and still Reflect.set on it. We should probably do that the `ls` operation too later.

zeeshanlakhani: right, I knew not the direct hamt itself, but the object w/ key "hamt" felt weird.

wnfs-wasm/src/fs/utils.rs
```diff
@@ -9,12 +9,19 @@ use wnfs::{
...
 //--------------------------------------------------------------------------------------------------
 
+pub(crate) fn error<E>(message: &str) -> impl FnOnce(E) -> js_sys::Error + '_
+where
+    E: Debug,
+{
+    move |e| Error::new(&format!("{message}: {e:?}"))
+}
+
 pub(crate) fn map_to_rust_vec<T, F: FnMut(JsValue) -> JsResult<T>>(
```
zeeshanlakhani, line 25: minor: this is not a rust vec at the end though right?

zeeshanlakhani: nvm, pr didn't give me the proper lines haha.

wnfs/src/common/utils.rs
```diff
@@ -84,6 +84,72 @@ pub fn get_random_bytes<const N: usize>(rng: &mut impl RngCore) -> [u8; N] {
...
+//--------------------------------------------------------------------------------------------------
+// Macros
+//--------------------------------------------------------------------------------------------------
+
+#[cfg(test)]
+pub(crate) mod test_setup {
+    #![allow(unused_macros)]
+    #![allow(unused_imports)]
+
+    macro_rules! init {
```
zeeshanlakhani, line 96: minor: comments on this macro

wnfs/src/private/file.rs
```diff
@@ -1,11 +1,27 @@
...
 
-use crate::{dagcbor, Id, Metadata, NodeType};
+//--------------------------------------------------------------------------------------------------
+// Constants
+//--------------------------------------------------------------------------------------------------
 
-use super::{namefilter::Namefilter, Key, PrivateNodeHeader, RevisionKey};
+/// The maximum block size is 2 ^ 18 but the first 12 bytes are reserved for the cipher text's initialization vector.
+/// This leaves a maximum of (2 ^ 18) - 12 = 262,132 bytes for the actual data.
+pub const MAX_BLOCK_CONTENT_SIZE: usize = MAX_BLOCK_SIZE - NONCE_SIZE;
```
zeeshanlakhani, line 24: minor: any good link to this in the spec?

zeeshanlakhani (approved):
minor comments left, great work.

wnfs-wasm/src/fs/utils.rs
appcypher: It is a rust vec in the end wrapped in Result.

zeeshanlakhani: oh i know. i had to go into the file and see what the underlying code was. The PR didn't help on the view haha.

PR #90, opened 2022-11-16
feat: private backpointer

matheus23:
Private backpointers as according to the spec: https://github.com/wnfs-wg/spec/blob/main/spec/private-wnfs.md#313-node-previous-backlinks

TLDR:
```diff
type PrivateDirectory = {
  type: "wnfs/priv/dir"
  version: "0.2.0"
  // encrypted using deriveKey(ratchet)
  header: Encrypted<CBOR<PrivateNodeHeader>>
+++  // encrypted using deriveKey(previousRatchet) where inc(previousRatchet) = ratchet
+++  previous: Encrypted<CBOR<Array<CID>>>

  // USERLAND
  ...
}
```

Three critical points in the code:
1. The places where we previously called `advance_ratchet`, we now need to do something extra: Before we advance the ratchet, we need to "have remembered" the CID that the directory/node was stored as and add that to the previous links *before* we advance the ratchet (so we can use the 'previous' revision key).
   This is mostly the `fix_up_path_nodes` function in `PrivateDirectory`.
2. `put` and `get` in the `PrivateForest`. This is where we last knew the `Cid` of node ciphertexts. Instead of throwing away the knowledge of the specific CID, we need to keep around this info (so we can use it in the places referred to in (1)).
   Effectively they get threaded through via `PrivateNode::load` and all the `deserialize_from_cbor` functions as the `from_cid: Cid` parameter.
3. In the previous iterators in `private/previous.rs` we need to resolve the bias passed to `PrivateForest::get` according to the decrypted previous links. I.e. when we look at the previous revision and there's multiple CIDs in the multivalue, just take the CID that matches one of the 'previous' CIDs.

TODO:
- [x] ~~Some code deduplication from multiple `store` implementations.~~ I think that's more involved and goes past only the `.store` implementation. `PrivateFile` and `PrivateDirectory` share a lot of common fields - they need to be refactored so the commonalities live together. May even turn into a spec change, as that'd may steer implementers into the right direction.
- [x] Everything I described in (3)
- [x] Rebasing / merging


matheus23:
I've decided to mark this as ready for review, intended to merge as-is, even though it *does not* actually fix history interleaving anomalies in the general case.
The reasoning is this:
- This is part of fixing history interleaving anomalies. Even if it's not the *whole* story.
- We'd rather merge this in and maintain it than pick it up later to create an even bigger PR and more stuff to rebase.

I've also fixed two issues in this PR, one of them would otherwise need more rebasing if they were extracted out from this PR:
- When moving a file, we need to copy its shards over, because the bare namefilter changed.
- Shard labels should be saturated namefilters.

wnfs/src/private/directory.rs
```diff
@@ -287,23 +291,64 @@ impl PrivateDirectory {
...
+
+    /// This prepares this file for key rotation, usually for moving or
+    /// copying the file to some other place.
+    ///
+    /// Will reset the ratchet, so a different key is necessary for read access,
+    /// will reset the inumber to reset write access,
+    /// will update the bare namefilter to match the new parent's namefilter,
+    /// so it inherits the write access rules from the new parent and
+    /// resets the `persisted_as` pointer.
+    pub(crate) fn prepare_key_rotation(
```
appcypher, line 326: How do we handle deterministic keys here? Is there a need to expose an alternative `attach` API for allowing deterministic key rotation?

wnfs/src/private/directory.rs
```diff
@@ -1251,15 +1302,66 @@ impl PrivateDirectory {
...
             header: {
                 let cbor_bytes = key.0.decrypt(&header).map_err(DeError::custom)?;
                 dagcbor::decode(&cbor_bytes).map_err(DeError::custom)?
             },
+            previous,
             entries,
         })
     }
+
+    pub async fn store<B: BlockStore>(&self, store: &mut B, rng: &mut impl RngCore) -> Result<Cid> {
```
appcypher, line 1317: Nit: Thinking we should stick to a convention here.

```rs
pub async fn store<B: BlockStore, R: RngCore>(&self, store: &mut B, rng: &mut R) -> Result<Cid>;
```

```rs
pub async fn store<B, R>(&self, store: &mut B, rng: &mut R) -> Result<Cid> where B: BlockStore, R: RngCore;
```

wnfs/src/private/encrypted.rs
```diff
@@ -0,0 +1,89 @@
...
+use super::Key;
+
+#[derive(Debug, Clone, Eq)]
+pub struct Encrypted<T> {
+    ciphertext: Vec<u8>,
+    value_cache: OnceCell<T>,
+}
+
+impl<T> Encrypted<T> {
+    pub fn from_value(value: T, key: &Key, rng: &mut impl RngCore) -> Result<Self>
```
appcypher, line 20: Do you want to add some docs here and in other public interface? The eager encryption is important to know.

wnfs/src/private/file.rs
```diff
@@ -405,10 +415,65 @@ impl PrivateFile {
...
+    /// This prepares this file for key rotation, usually for moving or
+    /// copying the file to some other place.
+    ///
+    /// Will reset the ratchet, so a different key is necessary for read access,
+    /// will reset the inumber to reset write access,
+    /// will update the bare namefilter to match the new parent's namefilter,
+    /// so it inherits the write access rules from the new parent and
+    /// resets the `persisted_as` pointer.
+    /// Will copy and re-encrypt all external content.
+    pub(crate) async fn prepare_key_rotation<B: BlockStore>(
```
appcypher, line 456: Same convention nit here

wnfs/src/private/file.rs
```diff
@@ -405,10 +415,65 @@ impl PrivateFile {
...
         label
     }
 
+    /// This should be called to prepare a node for modifications,
+    /// if it's meant to be a successor revision of the current revision.
+    ///
+    /// It will store the current revision in the given `BlockStore` to
+    /// retrieve its CID and put that into the `previous` links,
+    /// as well as advancing the ratchet and resetting the `persisted_as` pointer.
+    pub(crate) async fn prepare_next_revision<B: BlockStore>(
```
appcypher, line 429: Same convention nit here

wnfs/src/private/directory.rs
matheus23: That's a good question.

Generally, there's not *really* such a thing as "deterministic key rotation". When you rotate keys, you want to make sure that anyone who could read *previous* versions, won't be able to read *future* versions.
If you make that process deterministic however, then anyone who knows the past will be able to just apply the deterministic operation themselves to be able to read the present/future.

That said, one could argue our `Ratchet` is sort-of "deterministic key rotation": We constantly 'rotate' the key for every revision.

Anyway.

I think once we have shared private files, we should talk to Functionland again. Perhaps it's better to not generate `PrivateDirectory` and `PrivateRef` deterministically, but instead generate a single exchange public key deterministically, and 'share' read access of the public directory with that exchange key.

wnfs/src/private/directory.rs
matheus23: Generally I think the fewer type variables the better.
What do you think of

```rs
pub async fn store(&self, store: &mut impl BlockStore, rng: &mut impl RngCore) -> Result<Cid> {
```

?

matheus23:
Tried to address the style nits around passing `BlockStore` and `RngCore` traits by making most uses of them `impl BlockStore` and `impl RngCore`.

This worked well for `RngCore`, but there's some cases where we use `B: BlockStore + ?Sized`, and that doesn't translate nicely to `&mut (impl BlockStore + ?Sized)`.

Do we want to keep the `?Sized` marker? It seems like it really doesn't hurt and only enables *more* things with in practice. Why is the `Sized` assumption added implicitly in all generic parameters in rust then, though?

appcypher (approved):

PR #91, opened 2022-11-24
Allow user-provided ratchet seed and inumber

### LINKED ISSUE #86: exposing a constructor for PrivateRef
NB: Feature requests will only be considered if they solve a pain

# Summary: Exposing a constructor for PrivateRef

## Problem
The only way to obtain the root private key of WNFS-RS and use it on antoher device(login), right now, is to call .get_private_ref() on the directory on the first device and then serialize it and move it over to the other device. This doesn't allow deriving it deterministically.
In both the current version of fs used in webnative and WNFS go there is a constructor interface available to the programmer that can feed the private key that WNFS uses for the root. However, it is not exposed in WNFS rust version. exposing it allows application developer to feed the private key from their own login process (whatever it is) and it is deterministic,  and based on users' login and not random.

### Impact

We can then decrypt WNFS tree on any device that the user has without needing to transfer the keys as long as users logs in on different devices and the program can create the same keypair. It gives more flexibility to the application developer on what approach they want and also gives the user more control over which private key to use. If developer wants to feed a deterministic key, they can, if they want a random key, they still can do it.

## Solution

Exposing the constructor for PrivateRef

# Detail

**Is your feature request related to a problem? Please describe.**
It's frustrating where the login cannot be linked to WNFS encryption and we still need to transfer the private key between multiple devices even when the user completes a login process.

**Describe the solution you'd like**
Like webnative js version and go version, I want to be able to construct the Private_ref myself from a keypair or seed hash (both work).

**Describe alternatives you've considered**
The only way to obtain the root private key of WNFS-RS and use it on antoher device(login), right now, is to call .get_private_ref() on the directory on the first device and then serialize it and move it over to the other device. This doesn't allow deriving it deterministically.

**Additional context**
It is also discussed on IPFS Discord on #wnfs with Matheus and Boris

appcypher:
## Summary

In some cases, we want to be able to deterministically create a directory with seeds generated at userland, for example, from private keys. There is a hacky solution to this by passing a carefully designed rng, but that is likely to break in future versions.

This PR adds the API for accepting user-specified ratchet seed and inumber that solves the problem outlined above. 

- [x] Support user-specified ratchet seed and inumber
- [x] Expose corresponding wasm/js API
- [x] Bump version

## Test plan (required)
-  Testing
    
    ```bash
    rs-wnfs test --all
    ```

## Closing issues

Fixes #86

### DIFF
--- a/wnfs-wasm/Cargo.toml
+++ b/wnfs-wasm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasm-wnfs"
-version = "0.1.9"
+version = "0.1.10"
 description = "WebNative Filesystem API (WebAssembly)"
 keywords = ["wnfs", "wasm", "webnative", "ipfs", "decentralisation"]
 categories = [
--- a/wnfs-wasm/src/fs/private/directory.rs
+++ b/wnfs-wasm/src/fs/private/directory.rs
@@ -6,7 +6,10 @@ use chrono::{DateTime, Utc};
 use js_sys::{Array, Date, Promise, Uint8Array};
 use wasm_bindgen::{prelude::wasm_bindgen, JsValue};
 use wasm_bindgen_futures::future_to_promise;
-use wnfs::{Id, PrivateDirectory as WnfsPrivateDirectory, PrivateOpResult as WnfsPrivateOpResult};
+use wnfs::{
+    Id, PrivateDirectory as WnfsPrivateDirectory, PrivateOpResult as WnfsPrivateOpResult,
+    HASH_BYTE_SIZE,
+};
 
 use crate::{
     fs::{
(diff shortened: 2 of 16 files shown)


matheus23 (commented):
Nice! Should we also add a matching function `PrivateRef::from_seed(parent_bare_name: Namefilter, ratchet_seed: HashOutput, inumber: HashOutput) -> PrivateRef`?

matheus23:
What's the reasoning on using `with_*` naming?
I actually find it slightly confusing and think `from_*` makes more sense, because the parameters to the `from_seed` and `from_ratchet_key` functions deterministically define the `PrivateRef`. Every property in the `PrivateRef` depends on the arguments in these functions.
`with` seems to indicate there's 'something else' to it.

That said, I may just not know the normal rust conventions yet! :sweat_smile:

appcypher:
I don't mind either really but generally `with_*` seems to be preferred if you constructing an object out of other thing(s), and `from` if you are going to implement the `From<T>` trait. In this case that means implementing a `From<(Namefilter, HashOutput, ...)>` which doesn't feel nice. So I went with `with_*`. But I agree with you `from_*` sounds better.

Maybe this is sth @zeeshanlakhani can help out with.

matheus23:
Ah and @appcypher can you add a unit test that creates a `PrivateDirectory::with_seed`, stores it in a `PrivateForest`, then loads it again via a `PrivateRef::with_seed` with the same parameters? :pray:

matheus23 (approved):
Awesome!

PR #96, opened 2022-12-02
chore: put release-please process in place

zeeshanlakhani:
Includes:
    * pre-commit update for conventional commits
    * dependabot addition for deps
    * reverts *.toml v0.1.10 so that we can use the automated process on the next `fix`
      commit
    * won't publish bench and 0's out its version


.github/workflows/release.yml
```diff
@@ -0,0 +1,118 @@
...
+
+      - name: Verify Publishing of crate
+        uses: katyo/publish-crates@v1
+        with:
+          dry-run: true
+
+      - name: Cargo Publish to crates.io
+        uses: katyo/publish-crates@v1
+        with:
+          registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
zeeshanlakhani, line 68: @walkah @appcypher @matheus23, we need to set this in secrets. For publishing, it's prob set to run w/ @matheus23's token, but we should solve this once and for all.

.github/workflows/release.yml
```diff
@@ -0,0 +1,118 @@
...
+      - name: Build
+        run: |
+          wasm-pack build --target web --out-name wnfs
+          sed -i -e 's/"name": "wasm-wnfs"/"name": "wnfs",\n  "type": "module"/g' pkg/package.json
+
+      - name: Publish to npm
+        run: wasm-pack publish
+        working-directory: pkg
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
zeeshanlakhani, line 118: same as above: @walkah @appcypher @matheus23, we need to set this in secrets. For publishing, it's prob set to run w/ @matheus23's token, but we should solve this once and for all, though npm prob allows for multiple owners.

.github/workflows/release.yml
matheus23: I think it's only the `skip-ratchet` crate that I own. I don't think this is my token.

I see that there's a `crates-token` in our 1password, but which account is that associated with? There's a `fission-suite/Development` owner on the `wnfs` crate, but I can't add `fission-suite/Development` as an owner to `skip-ratchet`.

Anyway. What I wrote was mostly about `skip-ratchet`. For `wnfs` I think we should be able to just switch out the secret here and it should work, right? Since we already have `fission-suite/Development` as one of the owners of the `wnfs` crate.

.github/workflows/release.yml
matheus23: Looking at npm, the npmjs "developers team" of the fission org has write access. Do we have an NPM auth token for that team? (Is that how that works? :P )

.github/workflows/release.yml
zeeshanlakhani: @matheus23 yeah, let's hope :). @walkah would know best.

appcypher: I have a token but it is a personal access token. We need the `fission-suite/Development` token and there is an org-level `CARGO_REGISTRY_TOKEN` secret set that I would expect to work.

.github/workflows/release.yml
```diff
@@ -0,0 +1,118 @@
...
+
+      - name: Install wasm-pack
+        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+
+      # Builds output via wasm-pack and renames package.json name to remove
+      # `-wasm`.
+      - name: Build
+        run: |
+          wasm-pack build --target web --out-name wnfs
+          sed -i -e 's/"name": "wasm-wnfs"/"name": "wnfs",\n  "type": "module"/g' pkg/package.json
```
appcypher, line 112: Can this call the script instead so we don't miss future changes to the script.

.github/workflows/release.yml
walkah: I've added an org-level secret for `NPM_TOKEN` so this should work.

.github/workflows/release.yml
zeeshanlakhani: yeah, there are definitely not team tokens, but if it's from the owner, we're good.

appcypher (approved):
<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

PR #119, opened 2022-12-07
chore: fix-release flow

zeeshanlakhani:


.github/workflows/release.yml
```diff
@@ -23,7 +23,8 @@ jobs:
     if: github.ref == 'refs/heads/main' && github.repository_owner == 'wnfs-wg'
 
     outputs:
-      releases_created: ${{ steps.release.outputs['wnfs--release-created'] || steps.release.outputs['wnfs-wasm--release-created'] }}
+      releases_created: ${{ steps.release.outputs['wnfs--release_created'] || steps.release.outputs['wnfs-wasm--release_created'] }}
```
matheus23, line 26: I don't know what exactly is in scope here.
There seems to be `steps.release` and `needs.release-please`?
I'm just assuming `steps.release` doesn't need to be renamed to `steps.release-please` :D

matheus23 (approved):

.github/workflows/release.yml
zeeshanlakhani: right. this is just saying: "from the release (id) step, grab this item from the array"

PR #138, opened 2023-01-11
feat(api): self lookup & store at construction

appcypher:
## Summary

This PR implements the following **bugs/features**

- [x] Storing dirs in private forest at construction time
- [x] Self lookup of dirs to get the latest revision

Right now because of how PrivateDirectory constructor works, they don't store the dir right after creating. So the initial revision can be lost in that sense. Unlike `mkdir`, `write`methods which store the things you create in the forest during the operation. We are working on fixing that. This PR adds a fix is that puts the directory in the forest right after creating it.

In addition to that, we don't have a way to search_latest on the root directory. This PR adds support for self lookup by supplying an empty path segment to the lookup function.

## Test plan (required)
-  Testing
    
    ```bash
    rs-wnfs test --all
    ```

### DIFF
--- a/wnfs/examples/privateref.rs
+++ b/wnfs/examples/privateref.rs
@@ -20,11 +20,11 @@ async fn main() -> anyhow::Result<()> {
     // Some existing user key.
     let some_key = Key::new(utils::get_random_bytes::<32>(rng));
 
-    // Creating ratchet_seed from our key. And intializing the inumber and namefilter.
+    // Creating ratchet_seed from our user key. And intializing the inumber and namefilter.
     let ratchet_seed = Sha3_256::hash(&some_key.as_bytes());
     let inumber = utils::get_random_bytes::<32>(rng); // Needs to be random
 
-    // Create the directory from the ratchet_seed, inumber and namefilter. And save in forest.
+    // Create a root directory from the ratchet_seed, inumber and namefilter. Directory gets saved in forest.
     let PrivateOpResult {
         forest, root_dir, ..
     } = PrivateDirectory::with_seed_and_store(
@@ -53,7 +53,7 @@ async fn main() -> anyhow::Result<()> {
         )
         .await?;
 
-    // We can create a revision_key from the ratchet_seed.
+    // We can create a revision_key from our ratchet_seed.
     let ratchet = Ratchet::zero(ratchet_seed);
     let revision_key = RevisionKey::from(Key::new(ratchet.derive_key()));
 
@@ -64,7 +64,7 @@ async fn main() -> anyhow::Result<()> {
         rng,
     )?)?;
 
-    // We can deserialize the private_ref using the information we have.
+    // We can deserialize the private_ref using the revision_key at hand.
     let private_ref = decode_ipld(cbor, &revision_key)?;
 
     // Now we can fetch the directory from the forest.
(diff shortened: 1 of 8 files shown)


wnfs/src/private/directory.rs
```diff
@@ -143,6 +144,74 @@ impl PrivateDirectory {
         }
     }
 
+    pub async fn with_store<B: BlockStore, R: RngCore>(
```
matheus23, line 147: This tripped me up when reading through it. `with_store` sounds like you're constructing a `PrivateDirectory` and it'll contain a store afterwards.

Can we call this `new_and_store` perhaps? Or `new_with_store`?

For `with_seed_and_store`, should it be `new_with_seed_and_store`? In this case to me "with seed" makes sense, since the seed is used as input for creating the `PrivateDirectory`.

wnfs/src/private/directory.rs
```diff
@@ -728,6 +800,21 @@ impl PrivateDirectory {
...
+                    store,
+                )
+                .await?;
+
+            return Ok(match (search_latest, private_node) {
+                (true, Some(node)) => Some(node.search_latest(forest, store).await?),
+                (_, node) => node,
+            });
+        }
+
```
matheus23, line 817: Shouldn't we just have & export a function in `PrivateDirectory` that is called `search_latest`?
Using the empty string here as an indicator seems a little like a hack.

wnfs/src/private/node.rs
```diff
@@ -607,6 +607,21 @@ impl PrivateNodeHeader {
     }
 }
 
+impl Debug for PrivateNodeHeader {
```
matheus23, line 610: Do we want to keep custom `Debug` implementations?
I'm really not against it. I've previously had a bunch of these implementations for a couple of our structs in one of my PRs, but removed them since I felt like customizing them may be less ... idiomatic?

wnfs/src/private/directory.rs
appcypher: I agree we need a better name for those

wnfs/src/private/node.rs
appcypher: We want to keep them. The inumbers are easier read formatted as a single byte string esp. when pretty printing `{:#?}` is applied.

wnfs/src/private/directory.rs
appcypher: I thought it was bit hacky but wanted your input. Thinking there might be situations where that might be useful like with `cd`, `ls` commands. But I agree we should go for exposing `search_latest` for now.

matheus23: For `cd` and `ls`, they should just support handling empty paths then IMO.

wnfs/examples/privateref.rs
```diff
@@ -10,13 +10,16 @@ use wnfs::{
     utils, Hasher, MemoryBlockStore, Namefilter, PrivateDirectory, PrivateOpResult,
 };
 
-#[async_std::main]
+#[tokio::main(flavor = "multi_thread")]
```
matheus23, line 13: Any particular reason? Or mainly acknowledging that it's the more 'common' async runtime today?

wnfs/examples/privateref.rs
```diff
@@ -82,14 +91,11 @@ async fn main() -> anyhow::Result<()> {
...
-            .await?
-            .result
-            .unwrap()
-            .as_dir()?
-    };
+    let fetched_dir = fetched_node
+        .unwrap()
+        .search_latest(&forest, store)
+        .await?
+        .as_dir()?;
```
matheus23, line 98: :sparkles:

matheus23 (approved):
Clean!
Happy Evoli GIF

wnfs/examples/privateref.rs
appcypher: Just thought to try another runtime.

PR #144, opened 2023-01-16
feat(api): adds missing metadata functions for the private side

icidasset:
Adds `get_metadata` function to `PrivateFile` in `wnfs` crate and `metadata` function for the private file and directories. Also adds the `as_node` function to the private directories in WASM.


matheus23 (approved):
Check

I wonder why CI isn't running on these commits :eyes:

Issue #133, opened 2023-01-09
Implementation for Shared Private Files


matheus23:
Implementation of spec: https://github.com/wnfs-wg/spec/pull/51 (ready for review right now, but hopefully merged soon)

PR #148, opened 2023-01-17
feat(private): shared private data

appcypher:
## Summary

- [x] Sharing private data asynchronously
- [x] Receiving shared data

TODO
- [x] Proptests and unittests
- [x] wasm/js bindings

This PR adds support for the WNFS Shared Private Data Extension Specification protocol, that allows users to exchange private data asynchronously when the other party may be offline using store-and-forward networks. 

The protocol uses asymmetric encryption, with RSA public keys, and is versioned to support multiple versions on the same file system. Public keys are widely distributed in a "exchange keys partition" and are grouped by device for easy selection by the sender. 

Payloads containing the private data are stored in the "Private Forest" and are labeled with a name filter that includes the sender's and recipient's information, as well as a counter.

## Test plan (required)
-  Testing
    
    ```bash
    rs-wnfs test --all
    ```

Fixes https://github.com/wnfs-wg/rs-wnfs/issues/133


wnfs/src/private/share.rs
```diff
@@ -0,0 +1,508 @@
...
+        let payload = if temporal {
+            Self::Temporal(TemporalSharePointer {
+                label: private_ref.saturated_name_hash,
+                revision_key,
+            })
+        } else {
+            Self::Snapshot(SnapshotSharePointer {
+                label: private_ref.saturated_name_hash,
+                content_key: revision_key.derive_content_key(),
+            })
```
matheus23, line 176: Couldn't this function re-use functionality from `TemporalSharePointer::from_node` and `SnapshotSharePointer::from_node`?

wnfs/src/private/share.rs
```diff
@@ -0,0 +1,508 @@
...
+                        root_dir
+                            .lookup_node(&name, recipient_store)
+                            .await?
+                            .ok_or(FsError::NotFound)?
+                            .as_file()?;
+
+                        let PublicOpResult { result: cid, root_dir: root } = root_dir.read(&[name], recipient_store).await?;
+                        root_dir = root;
+                        yield recipient_store.get_block(&cid).await?.to_vec();
+                        break
```
matheus23, line 384: Is this `break` necessary?

matheus23 (commented):
This is pretty good :)

Apart from the comments, something more general: I think we should have another primitive that allows us to go from `(recipientExchangeKey, senderDID) -> count`, so we can "opportunistically look for share payloads". This way we know what to initialize the count with.
The idea would be that such a function simply looks up which private forest labels are already occupied and it jumps to the first unoccupied one.

wnfs/src/private/share.rs
appcypher: That was the plan. Somehow forgot to do that

wnfs/src/private/share.rs
appcypher: Yes. It breaks the inner loop cos the yield won't. There is however a bug in this impl and already fixed that

wnfs/src/private/keys/rsa.rs
```diff
@@ -0,0 +1,111 @@
...
+    where
+        Self: Sized;
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+#[derive(Debug, Clone)]
+pub struct RsaKeys {
+    pub(crate) public_key: RsaPublicKey,
+    pub(crate) private_key: Option<RsaPrivateKey>,
+}
```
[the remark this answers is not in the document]
appcypher: Interesting.

wnfs/src/private/node.rs
```diff
@@ -48,11 +48,18 @@ pub enum PrivateNode {
...
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
-pub struct RevisionKey(pub Key);
+pub struct RevisionKey(pub AesKey);
+
+/// The two types of keys.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum KeyType {
+    Temporal(ContentKey),
+    Snapshot(RevisionKey),
+}
```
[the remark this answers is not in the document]
appcypher: Getting rid of the enum. It is not longer used in the main impl.

wnfs/src/private/share.rs
```diff
@@ -0,0 +1,629 @@
...
+                .unwrap();
+
+        // Assert payload is the same as the original.
+        assert_eq!(node.as_dir().unwrap(), sharer_dir);
+    }
+
+    #[async_std::test]
+    async fn serialized_share_payload_can_be_deserialized() {
+        async_std::task::block_on(async {
+            let (forest, store, rng) = test_setup::init!(forest, mut store, mut rng);
```
matheus23, line 610:
```suggestion
    #[async_std::test]
    async fn serialized_share_payload_can_be_deserialized() {
        let (forest, store, rng) = test_setup::init!(forest, mut store, mut rng);
```

wnfs/Cargo.toml
```diff
@@ -28,6 +28,7 @@ chrono = { version = "0.4.23", default-features = false, features = ["clock", "s
 either = "1.8"
 futures = "0.3"
 futures-util = "0.3"
+hex = "0.4"
```
matheus23, line 31: Stray dependency?

```suggestion
```

matheus23 (commented):
Some small nits.

Can you explain why you've made the recipient's and sender's `BlockStore`s separate & stored within the structs?
So far, we've been working with `&mut impl BlockStore` being passed. That allows a caller to choose the block store they want to use per-operation.
And storing the `BlockStore` within the `Sender` and `Recipient` forces the caller to separate these `BlockStore`s: Try unifying `recipient_store` and `sender_store` in the test.

---

Apart form that, once this PR has a function that computes a `share_count`, I think this is good to go :)

wnfs-wasm/tests/share.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+          MemoryBlockStore,
+          Rng,
+          createSharerDir,
+          createRecipientExchangeRoot,
+          ExchangeKey,
+        },
+      } = await window.setup();
+
+      // @ts-ignore
+      globalThis.ExchangeKey = ExchangeKey;
```
matheus23, line 32: Yeah. This goes on the pile of things I'm not happy about with wasm-bindgen's choice of not exposing an `exports = await init(imports)` kind of api :/

wnfs-wasm/tests/share.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+        sharerDir.asNode(),
+        true,
+        sharerForest,
+        sharerStore,
+        rng
+      );
+
+      var sharerForest2 = await share(
+        sharerPayload,
+        BigInt(0),
```
matheus23, line 63: Can we make this not rely on `BigInt`?
We've found that lots of tools still can't deal with them. Using just 32 bits should be totally fine. I expect nobody to share more than 4 billion things with another person :D

matheus23 (approved):
This is looking good except for a minor nit thing with `BigInt` vs. using normal numbers in JS. (I think that's just a `u64` vs. `u32` thing on the rust side).

PR #149, opened 2023-01-22
refactor: Double random

pinkforest:
## Summary

Re-factors double_Random and adds `#![deny(unsafe_code)]`

The old one used [transmute](https://doc.rust-lang.org/stable/std/mem/fn.transmute.html)

## Test plan (required)

I benched with two Rng's just in case and it's ~same and added a test

I even implemented the stdlib safer version (with little unsafe) for comparison in case it is still needed for some reason

https://github.com/pinkforest/double-random

### DIFF
--- a/wnfs/src/private/node.rs
+++ b/wnfs/src/private/node.rs
@@ -133,13 +133,6 @@ impl PrivateNode {
         }
     }
 
-    pub(crate) fn generate_double_random(rng: &mut impl RngCore) -> (HashOutput, HashOutput) {
-        (
-            crate::utils::get_random_bytes::<HASH_BYTE_SIZE>(rng),
-            crate::utils::get_random_bytes::<HASH_BYTE_SIZE>(rng),
-        )
-    }
-
     /// Updates bare name ancestry of private sub tree.
     #[async_recursion(?Send)]
     pub(crate) async fn update_ancestry(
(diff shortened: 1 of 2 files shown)


wnfs/src/private/node.rs
```diff
@@ -714,6 +710,13 @@ mod tests {
 
     use super::*;
 
+    #[async_std::test]
+    async fn generate_double_random_is_ne() {
+        let rng = &mut TestRng::deterministic_rng(RngAlgorithm::ChaCha);
+        let (first, second) = PrivateNode::generate_double_random(rng);
+        assert_ne!(first, second);
+    }
```
appcypher, line 718: Thanks for the refactor. 
Thinking this can work as a non-async function.

```suggestion
    #[test]
    fn generate_double_random_is_ne() {
        let rng = &mut TestRng::deterministic_rng(RngAlgorithm::ChaCha);
        let (first, second) = PrivateNode::generate_double_random(rng);
        assert_ne!(first, second);
    }
```

matheus23:
Shouldn't we perhaps just push the `utils::get_random_bytes` calls to the use-sites and skip the `generate_double_random` calls?

In any case, I'd :+1: this PR already after the test chnage.

pinkforest:
> Shouldn't we perhaps just push the utils::get_random_bytes calls to the use-sites

Yup agreed - thanks :tada:

wnfs/src/private/node.rs
pinkforest: Thanks - Re-wired the calls directly to util as it was single-use fn

matheus23 (approved):
Woman asking 'is this safe?'

Yes it is!

PR #151, opened 2023-01-23
chore: use `#!/usr/bin/env bash` instead of `#!/bin/bash`

matheus23:
This basically fixes the script for me in NixOS, where there's no such file `/bin/bash`. 
This should work on Mac, too.

`/usr/bin/env <name>` just searches for `<name>` in your `$PATH` and executes that.


appcypher (approved):
LGTM! :tada:

Issue #153, opened 2023-01-25
wasm-bindgen interfaces for working with `PrivateRef`s


matheus23:
Internally, we're making use of `PrivateRef`. They're likely a good abstraction (and we've used it internally), so we should also have them in the wnfs-wasm interface.

matheus23:
This is implemented as of #159

PR #159, opened 2023-02-03
Implement non-nested encryption

matheus23:
Possible todos before merging:
- [x] rename `RevisionKey` into `TemporalKey` and
- [x] rename `ContentKey` into `SnapshotKey`?
- [x] resolve all `TODO(matheus23) docs`
- [X] ~~perhaps figure out a way to not have both `PrivateNode::store` and `PrivateForest::put`. However it's possible that'll be a next-PR thing when we introduce something like `PrivateLink`? Although needs though (esp. with `.store` usage in `prepare_next_revision`)~~ Yeah. It's a next-PR thing and I'm hoping we can solve this with `PrivateLink`.
- [x] Fix wasm API from refactor


matheus23:
Base branch for this is #148, so depends on that.

matheus23:
Closes #153

wnfs/src/private/privateref.rs
```diff
@@ -156,30 +167,67 @@ impl Debug for PrivateRef {
...
+    use futures::StreamExt;
 
-    use super::PrivateRef;
+    use super::RevisionRef;
 
     #[async_std::test]
-    async fn can_create_privateref_deterministically_with_user_provided_seeds() {
-        let (forest, store, rng) = test_setup::init!(forest, mut store, mut rng);
+    async fn can_create_revisionref_deterministically_with_user_provided_seeds() {
+        let (ref mut forest, store, rng) = test_setup::init!(forest, mut store, mut rng);
```
appcypher, line 230: Nit: I would like us to stick to a convention here. So we could use `ref mut` everywhere or we could support `mut forest` in the macro impl.

matheus23:
All merge conflicts should be resolved now & all resulting test errors fixed :)

wnfs-wasm/Cargo.toml
```diff
@@ -25,7 +25,7 @@ console_error_panic_hook = { version = "0.1", optional = true }
 getrandom = { version = "0.2", features = ["js"] }
 js-sys = { version = "0.3", optional = true }
 rand_core = "0.6"
-wasm-bindgen = { version = "0.2", optional = true, features = ["serde-serialize"] }
+wasm-bindgen = { version = "0.2.84", optional = true, features = ["serde-serialize"] }
```
appcypher, line 28: Does having the minor version affect anything?

wnfs-wasm/playwright.config.ts
```diff
@@ -49,18 +49,18 @@ const config: PlaywrightTestConfig = {
...
+    //   use: {
+    //     ...devices["Desktop Firefox"],
+    //   },
+    // },
+    // {
+    //   name: "safari",
+    //   use: {
+    //     ...devices["Desktop Safari"],
+    //   },
+    // },
```
appcypher, line 63: Forgot to add them back?

wnfs-wasm/src/fs/private/directory.rs
```diff
@@ -81,14 +81,13 @@ impl PrivateDirectory {
     ) -> JsResult<Promise> {
         let time = DateTime::<Utc>::from(time);
         let mut store = ForeignBlockStore(store);
+        let mut forest = Rc::clone(&forest.0);
```
appcypher, line 84: PrivateForest is owned. Can we do this instead?

```suggestion
        let mut forest = forest.0;
```

wnfs-wasm/src/fs/private/directory.rs
```diff
@@ -118,21 +117,21 @@ impl PrivateDirectory {
         let ratchet_seed = utils::expect_bytes::<HASH_BYTE_SIZE>(ratchet_seed)?;
         let inumber = utils::expect_bytes::<HASH_BYTE_SIZE>(inumber)?;
         let mut store = ForeignBlockStore(store);
+        let mut forest = Rc::clone(&forest.0);
```
appcypher, line 120: Same here
```suggestion
        let mut forest = forest.0;
```

wnfs-wasm/Cargo.toml
matheus23: Yes, there's some important bugfixes in 0.2.84: https://github.com/rustwasm/wasm-bindgen/pull/3117

wnfs-wasm/playwright.config.ts
matheus23: Yes. My bad :)

matheus23:
@appcypher and I shortly discussed on discord whether we should prefer taking owned or borrowed types in the bindgen API.
We agreed to take owned types for all types that are basically just wrappers around `Rc`. This way we only have a single layer of pointer indirection and less (no?) `Rc::clone` calls.

wnfs/src/private/privateref.rs
```diff
@@ -1,149 +1,160 @@
...
+/// A pointer to a specific revision in the private forest
+/// together with the TemporalKey to decrypt any of these
+/// revisions.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RevisionRef {
+    /// Sha3-256 hash of saturated namefilter. Used as the label for private nodes in the private forest.
+    pub saturated_name_hash: HashOutput,
+    /// Skip-ratchet-derived key. Gives read access to the revision pointed to and any newer revisions.
+    pub temporal_key: TemporalKey,
 }
```
appcypher, line 48: This is fine but thinking of an alternative way of going about this. As in:

```rust
pub struct Ref<T> {
    pub saturated_name_hash: HashOutput,
    pub temporal_key: TemporalKey,
    pub t: T,
}

pub type PrivatRef = Ref<Cid>;
pub type RevisionRef = Ref<()>;

impl PrivateRef { /* ... */ }
impl RevisionRef { /* ... */ }
```

appcypher:
Honestly this is great work! Tackles multiple issues together. Apart from the nitpicks above, I think this good to go.

wnfs/src/private/privateref.rs
matheus23: Yeah true. Another way to slice the problem would be this:
```rs
pub struct PrivateRef {
  revision_ref: RevisionRef,
  content_cid: Cid,
}
```

There's probably more ways. I like to keep things simpler at the cost of some duplication. Would that be ok for you, too?

(There's also perhaps some changes coming with the `PrivateLink` abstraction eventually that *may* throw a wrench into abstractions we'd introduce now)

appcypher (approved):
Looks good! 🎉

PR #163, opened 2023-02-08
feat: Streaming write for PrivateFile

### LINKED ISSUE #152: Streaming Write for `PrivateFile`
# Summary

Support for streaming writes to create new `PrivateFile`s, instead of having to buffer the whole file's contents in advance (by [providing a `Vec<u8>`](https://github.com/wnfs-wg/rs-wnfs/blob/703d1c193e5510d14652c97567f1b2f57b878d01/wnfs/src/private/file.rs#L177))

## Problem

The requirement to buffer means we need to allocate a lot of memory in advance unnecessarily.

### Impact

It may be the reason why existing integrations (e.g. FxPhotos) error out with files >~200Mb.

## Solution

Implement streaming writes somehow. Perhaps by accepting a `content: impl Stream<Item = Vec<u8>>` as parameter?

matheus23:
Closes #152 

This is a minimal effort PR for importing big data into WNFS without insane memory requirements by streaming it in during file creation.
The incoming stream is split into private file chunks and encrypted & serialized one-by-one. Each chunk gets written before the next chunk gets fetched, thus this should take only roughly constant memory.

In the future we should look two more things:
1. We want to provide better APIs in `PrivateDirectory::write`, where currently that only takes a `Vec<u8>` of bytes. We probably need to figure out something akin to `PrivateDirectory::open` which returns a `PrivateFile`, which can later be used in `PrivateDirectory::write` (perhaps keeping the old version, too, because it's a nice shortcut for small files).
2. The most flexible API will be something like a `PrivateFile` that implements `AsyncRead + AsyncWrite + AsyncSeek`. However, to make that performant on big files we need content-defined chunking.


matheus23:
Flipping this to ready for review.
That said, the merge order should be:
#148 
#159 
Then this.

wnfs-wasm/playwright.config.ts
```diff
@@ -49,18 +49,18 @@ const config: PlaywrightTestConfig = {
...
+    //   use: {
+    //     ...devices["Desktop Firefox"],
+    //   },
+    // },
+    // {
+    //   name: "safari",
+    //   use: {
+    //     ...devices["Desktop Safari"],
+    //   },
+    // },
```
appcypher, line 63: ☝🏾 Missed

wnfs/src/private/file.rs
```diff
@@ -668,6 +787,30 @@ mod tests {
             content[2 * MAX_BLOCK_CONTENT_SIZE..4 * MAX_BLOCK_CONTENT_SIZE]
         );
     }
+
+    #[async_std::test]
+    async fn can_construct_file_from_stream() {
+        let disk_file = File::open("./src/private/directory.rs").await.unwrap();
```
appcypher, line 793: Fixture? Source file likely to change.

wnfs/src/private/file.rs
```diff
@@ -364,6 +427,60 @@ impl PrivateFile {
...
+        let mut block_index = 0;
+
+        loop {
+            let mut current_block = vec![0u8; MAX_BLOCK_SIZE];
+            let nonce = SnapshotKey::generate_nonce(rng);
+            current_block[..NONCE_SIZE].copy_from_slice(&nonce);
+
+            // read up to MAX_BLOCK_CONTENT_SIZE content
+
+            let content_end = NONCE_SIZE + MAX_BLOCK_CONTENT_SIZE;
```
appcypher, line 452: Worth making into a constant?

wnfs/src/private/node.rs
```diff
@@ -821,14 +824,32 @@ impl SnapshotKey {
...
-        Ok([nonce_bytes.to_vec(), cipher_text].concat())
+        Ok([nonce.to_vec(), cipher_text].concat())
+    }
+
+    /// Generates a random 12-byte nonce for encryption.
+    pub(crate) fn generate_nonce(rng: &mut impl RngCore) -> Nonce<U12> {
+        let mut nonce = Nonce::default();
+        rng.fill_bytes(&mut nonce);
+        nonce
+    }
```
appcypher, line 841: 👍🏾

wnfs/src/common/utils.rs
```diff
@@ -66,6 +67,29 @@ pub(crate) fn split_last(path_segments: &[String]) -> Result<(&[String], &String
...
+            break;
+        }
+
+        if bytes_read == buffer.len() {
+            break;
+        }
+    }
+    Ok((bytes_read, done))
+}
+
```
appcypher, line 92: Interesting, `read_to_end` won't be sufficient because it can resize the input buffer.

wnfs/src/private/file.rs
matheus23: Yeah. You're right. I should just find some decent fixture.

wnfs/src/private/file.rs
matheus23: Hm, I don't think so. At least not a constant expose at the top level.

wnfs/src/common/utils.rs
matheus23: Yeah exactly. I wanted something with fewer copies.
I know the buffer size in advance (it's the file chunk size) and just fill it.
`read_to_end` just gives me *some* size, and I don't have any control over it.

matheus23:
For the file to test in the fixture, I went with a (public domain) recording of Clara Schumann's Scherzo No. 2. It's a classical music piece and Clara Schumann rocks (even though we've forgotten about her nowadays and her husband took the spotlight in today's history books).

Honestly, I'm up for other ideas though! If anyone knows some cool fixture file to use that is >1MB and <10MB, let me know.

appcypher (approved):
<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

PR #164, opened 2023-02-16
chore(bench): add hamt diff and merge benches

appcypher:


matheus23 (approved):
Measurements! Thanks :pray: 

GIF of dog helping holding a tape-measure

As we've already discussed live, having this in before #169 is cool since we hope to see a performance improvement in merges :)

PR #165, opened 2023-02-16
refactor(hamt): refactor key value diff

appcypher:
## Summary

- [x] Refactor key-value diff

This PR refactors key-value diff which initially uses node-diff resulting in not so efficient implementation that traverse the hamt multiple times. There was plan to refactor the merge impl as well to support modifying nodes in the hamt in place but I realised I'm just going to be using Node::set_value anyway. The gains there is not worth the complexity of the code.

Added benches in #164 

## Test plan (required)
-  Testing
    
    ```bash
    rs-wnfs test --all
    ```

Fixes https://github.com/wnfs-wg/rs-wnfs/issues/127


wnfs/src/private/hamt/diff/key_value.rs
```diff
@@ -66,105 +68,200 @@ type EitherPairOrNode<'a, K, V, H> = Option<Either<&'a Pair<K, V>, &'a Rc<Node<K
...
+                        })
+                    },
+                    store,
+                )
+                .await
+        }
+    }
+}
+
+async fn generate_modify_changes<K, V, H>(
```
matheus23, line 197: I didn't think the function name was 100% fitting, so here's a suggestion:

```suggestion
async fn pointers_kv_diff<K, V, H>(
```

That said, this is an internal helper method, so not super important.

wnfs/src/private/hamt/diff/node.rs
```diff
@@ -69,33 +69,30 @@ pub struct NodeChange {
 ///    println!("Changes {:#?}", changes);
 /// }
 /// ```
-#[async_recursion(?Send)]
-pub async fn node_diff<K, V, H, B>(
+pub async fn node_diff<K, V, H>(
```
matheus23, line 72: Is the node diff now actually used? IIUC, we don't use it for `kv_diff` and we're not using it for `merge` either, right?
I'm thinking we probably don't want to expose it, unless we'd know a specific use case for it? If we delete it now, it's not like it's gone, we can go back and grab it from the history it if it turns out to be useful, but removing it now will make the public API surface smaller & potentially reduce maintenance burden.

matheus23 (approved):
Except for potentially removing `node_diff` now, this is good to go by me! :+1: 

> There was plan to refactor the merge impl as well to support modifying nodes in the hamt in place but I realised I'm just going to be using Node::set_value anyway. The gains there is not worth the complexity of the code.

Let me preface saying this with "we probably want to split that out into another PR":
I wouldn't say the [original idea](https://github.com/wnfs-wg/rs-wnfs/pull/94#discussion_r1050911710) was modifying nodes in the hamt in place.
It was more about not requiring a call to `kv_diff`.
To illustrate what I was thinking: Imagine two different HAMTs. Their root `Node`s bitmaps are `1111_1111_0000_0000` and `0000_0000_1111_1111` respectively (let's ignore for a moment that this is very unlikely), but a merge implementation wouldn't even need to look at the two root node's children, it could just merge their `Pointers` together. This means it's short-circuiting a bunch. If instead it uses `kv_diff`, it needs to fetch what's behind the `Pointers` before merging.

matheus23 (approved):
Noice meme GIF

PR #167, opened 2023-02-18
Make re-exports more flexible

appcypher:


wnfs/src/lib.rs
```diff
@@ -26,20 +26,10 @@ pub use traits::*;
...
 
-pub mod rng {
+pub mod rand_core {
     pub use rand_core::RngCore;
 }
 
-pub mod ratchet {
-    pub use skip_ratchet::{Ratchet, RatchetSeeker};
-}
+pub use skip_ratchet;
```
matheus23, line 35: I've been thinking about these re-exports lately when I looked at the public API:
I'm on board with the `libipld` and `rand_core` exports, since they appear in `wnfs`'s public api. However, ideally a downstream user of `wnfs` wouldn't need to worry about knowing what a `skip_ratchet` is or does, so perhaps we shouldn't expose it?

matheus23 (approved):

PR #171, opened 2023-02-20
chore: Strip flags 2

pinkforest:
Supercedes https://github.com/wnfs-wg/rs-wnfs/pull/158#pullrequestreview-1306266384

Testing in CI


appcypher:
And as suggested by @pinkforest, it would be nice to have a build script give user different options based on their prefs.

PR #172, opened 2023-02-22
feat: PrivateLink abstraction

### LINKED ISSUE #58: Optimize PrivateForest Access
`PrivateForest` is currently defined as `pub type PrivateForest = Hamt<Namefilter, Cid>;` where Cid points to an encrypted deserialized `PrivateNode` bytes. For every access to this structure we have to decrypt the `PrivateNode` each time. This is inefficient so we want to cache the `PrivateNode`s in a `PrivateLink` type just like `PublicLink`.

I was working on this but soon found out current `AsynSerialize` definition is not enough to allow `PrivateLink` implementation. We should look into how we can make `AsyncSerialize` more robust or have a variant of it for the Private filesystem.

matheus23:
## Summary

This PR introduces an internal abstraction similar to the public file system's `Link` abstraction, but for the private side: The `PrivateLink`.
This makes it possible to work in a sort of "staging" area of the file system that is much faster: It doesn't require writing values to the `PrivateForest`, it doesn't deserialize or even encrypt during operation.
The file system flush then happens when the developer calls `.store`, which will recursively encrypt, serialize and store resulting ciphertexts in the `PrivateForest`.

Closes #58

## TODO

- [x] Consider also moving `forest.get` into `PrivateNode/PrivateFile/PrivateDirectory::load`
- [x] Remove `&mut Rc<...>` and other `&mut` requirements which can now become `&PrivateForest` and `&impl BlockStore`.


wnfs-wasm/src/fs/private/forest.rs
```diff
@@ -50,107 +31,37 @@ impl PrivateForest {
         Self(Rc::new(WnfsPrivateForest::default()))
     }
 
-    #[wasm_bindgen]
-    pub fn get(&self, private_ref: PrivateRef, store: BlockStore) -> JsResult<Promise> {
+    /// Loads an existing private forest from a given CID
+    /// You need to have previously `.store()`ed it to get its CID.
+    pub fn load(cid: Vec<u8>, store: BlockStore) -> JsResult<Promise> {
```
matheus23, line 36: This diff is horrible :sweat_smile: 

I've removed all previous functions from `PrivateForest`'s Wasm API (`put` and `get`).

Instead, I realized we still needed `store` and `load` for webnative, so these got added.

wnfs/src/private/directory.rs
```diff
@@ -47,15 +47,15 @@ pub type PrivatePathNodesResult = PathNodesResult<PrivateDirectory>;
...
 
 #[derive(Debug)]
 pub struct PrivateDirectoryContent {
     persisted_as: OnceCell<Cid>,
-    pub previous: BTreeSet<(usize, Encrypted<Cid>)>,
-    pub metadata: Metadata,
-    pub entries: BTreeMap<String, PrivateRef>,
+    pub(crate) previous: BTreeSet<(usize, Encrypted<Cid>)>,
+    pub(crate) metadata: Metadata,
+    pub(crate) entries: BTreeMap<String, PrivateLink>,
```
matheus23, line 58: It seemed like it'd make sense to make the `PrivateDirectoryContent` and related things `pub(crate)`.
`PrivateLink` isn't in the public API, so that'd leak otherwise.
Also, we've got stuff like `.get_metadata()`, so I felt it'd be weird if you could both do `.metadata` and `.get_metadata()`.

I'm somewhat unsure about `.header`, though. We use it in the examples, but I'm thinking perhaps we should eventually figure out a good API for developers to get what they want to get from the `header`. Technically I'd consider it an implementation detail that there's a separate header for directories (and I can see ways in which we'll change the inner data structures in the future to separate them differently).

wnfs/src/private/node.rs
```diff
@@ -451,14 +445,65 @@ impl PrivateNode {
...
+            .await?
+        {
+            Some(cids) if cids.contains(&private_ref.content_cid) => private_ref.content_cid,
+            _ => return Err(FsError::NotFound.into()),
+        };
+
+        Self::from_cid(cid, &private_ref.temporal_key, store).await
+    }
+
+    pub(crate) async fn from_cid(
```
matheus23, line 498: I renamed the old `PrivateNode::load` into `from_cid` essentially.
Other than that it doesn't take the whole `PrivateRef` anymore, but just the `Cid` and the `TemporalKey`.
That felt like it made sense, since it doesn't do any `PrivateForest` lookups anyway. (now explicitly!)

wnfs/examples/privateref.rs
```diff
@@ -64,10 +58,10 @@ async fn main() -> anyhow::Result<()> {
     // --------- Method 1: Exchange serialized revision ref -----------
 
     // serialize the root_dir's revision_ref.
-    let cbor = encode(&root_dir.header.derive_revision_ref())?;
+    let cbor = dagcbor::encode(&root_dir.header.derive_revision_ref())?;
```
appcypher, line 61: Nice. 👍

wnfs/src/private/file.rs
```diff
@@ -554,36 +554,45 @@ impl PrivateFile {
...
+    /// This doesn't have any effect if the current state hasn't been `.store()`ed yet.
+    /// Otherwise, it clones itself, stores its current CID in the previous links and
+    /// advances its ratchet.
+    pub(crate) fn prepare_next_revision(self: Rc<Self>) -> Result<Self> {
+        let previous_cid = match self.content.persisted_as.get() {
+            Some(cid) => *cid,
+            None => {
+                // The current revision wasn't written yet.
+                // There's no point in advancing the revision even further.
+                return Ok(Rc::try_unwrap(self).unwrap_or_else(|rc| (*rc).clone()));
```
appcypher, line 566: Notice we have this pattern in a couple places. Thinking we should bring `UnwrapOrClone` trait back.

matheus23: Yeah I know. The only reason I didn't bring it back was because we'll get rid of these as soon as we'll do the `&mut Rc<Directory>` stuff, since then the signature of this function should change to `(self: &mut Rc<Self>) -> &mut Self` and it makes use of `Rc::make_mut` instead.

appcypher (approved):
<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

<h3 align="center">This looks fantastic to me! 📈 </h3>

PR #174, opened 2023-02-22
feat: Remove `base_history_on` and auto-track history instead

matheus23:
This is following the work on `PrivateLink`, it makes the history-tracking and previous-link creation in the public file system work similarly to how it works in the private file system (after merging the `PrivateLink` PR).

Previously, we created previous links in the public file system via a function called `base_history_on`. It had some downsides:
- You need to keep track of the 'old' file system state to compare to. This sometimes causes refcells to have counts >1, which means we can't do as much (safe) in-place mutation.
- You walk the file system "twice" for each change, at least. Once to do the change, then a second time when you're looking through the file system for nodes that need to update their `previous` links.


matheus23:
We thought of making this part of #172, but decided having it separate for review is helpful.

So it's also *spiritually* related to #58

matheus23:
I think this PR is much more controversial / has room for improvement.
Good call on extracting it out.

Some points of contention:
* What should our philosophy regarding the `BlockStore` trait be? I've been moving more and more functionality out of it to use-sites. E.g. we're no more using `put_async_serializable`, and we've never used it on the private side, since we didn't have sufficient traits to abstract over what serialization needs on the private side. Similar questions could be asked with the `get_remembering_persistence` function (and all its "possible variants").
* I don't like that `RemembersPersistence` exposes the `persisted_as` OnceCell. Is there a way to implement the trait without exposing it in the API, but still expose it to `Link`?
* more stuff I may be forgetting.

wnfs/src/common/blockstore.rs
```diff
@@ -28,40 +27,17 @@ pub trait BlockStore {
         self.put_block(bytes, IpldCodec::DagCbor).await
     }
 
-    async fn put_private_serializable<V: Serialize>(
```
matheus23, line 31: I removed these. Since I added `SnapshotKey::encrypt` and `TemporalKey::key_wrap_encrypt`, we weren't making use of a single `put_private_serializable` anymore.
We *could* add two more functions for en/decrypting on the `BlockStore` trait, but it felt like it'd balloon in size at some point and have quite some dependencies.

wnfs/src/common/link.rs
```diff
@@ -215,37 +224,99 @@ where
     }
 }
 
+pub trait RemembersPersistence {
```
appcypher, line 227: Nit: Thinking there should be a mention of Cid in the name because that is what it is really about. Maybe like `CidPersistable`, `PersistAsCid`, etc.

matheus23: How about `RemembersCid`?

appcypher:
I'm trying to weigh the pros and the cons. As you have mentioned, the new `Link` implementation now places a bunch of requirements on T. T now has to persist its own cid and implement some traits. This makes sense for files and dirs but not so much for Node. On the bright side, the public file system now works closely like the private side in that in fixing the path nodes you are also fixing the revision.
My question now would be, isn't `baseHistoryOn` an important api for reaching feature parity with webnative?

appcypher (approved):

wnfs/src/common/link.rs
appcypher: Sounds fine

wnfs/src/common/link.rs
```diff
@@ -21,16 +24,17 @@ pub enum Link<T> {
     /// further calls to `resolve_value` will just return from that cache.
     Encoded { cid: Cid, value_cache: OnceCell<T> },
     /// A variant of `Link` that started out as a deserialized value `T`.
-    /// If the cid is resolved using `resolve_cid`, then the `cid_cache` gets populated and further calls
-    /// to `resolve_cid` will just return from that cache.
-    Decoded { value: T, cid_cache: OnceCell<Cid> },
+    /// If the cid is resolved using `resolve_cid`, then `T`'s `.persisted_as` from the
+    /// `RemembersPersistence` trait is called and that `OnceCell<Cid>` is populated, preventing
```
appcypher, line 28:
```suggestion
    /// `RemembersCid` trait is called and that `OnceCell<Cid>` is populated, preventing
```

matheus23:
> I'm trying to weigh the pros and the cons. As you have mentioned, the new Link implementation now places a bunch of requirements on T. T now has to persist its own cid and implement some traits. This makes sense for files and dirs but not so much for Node. On the bright side, the public file system now works closely like the private side in that in fixing the path nodes you are also fixing the revision.

Yeah, this is the downside. I was anxious about this at first, but luckily it wasn't *too crazy* of a change for `Node` to just have its own `OnceCell`, except that I couldn't derive `Clone` and so on anymore.

> My question now would be, isn't `baseHistoryOn` an important api for reaching feature parity with webnative?

No, luckily not. `webnative` doesn't have that, only the `wnfs2` branch that got turned into `rs-wnfs`, and even there, it wouldn't have been a public API in webnative.

It was a way of generating `previous` links, without *always* generating them when anything changes. Because if we were forced to *always* generate previous links, we would be forced to always serialize, too.

But this PR (and the same on the private side) is another way of achieving that.

Something related: We've been thinking something like `historyStep()`, which you could call on any node. That would force a previous link to be generated (think of it like `git commit --allow-empty`). I think we can do that by breaking out the previous-link-generating part out of `prepare_next_revision` in the future.
But, yeah, future! :)

PR #184, opened 2023-02-24
refactor: separate into crates

appcypher:
## Summary

This PR breaks the wnfs crate into separate smaller crates and attempts to fix some of import cycles in the project

This is necessary to keep the core fs implementation lean and isolate concerns.

## Test plan (required)

```bash 
scripts/rs-wnfs.sh test
```

## Closing issues

Closes https://github.com/wnfs-wg/rs-wnfs/issues/183


.release-please-manifest.json
```diff
@@ -1,5 +1,8 @@
 {
   "wnfs": "0.1.16",
-  "wnfs-wasm": "0.1.16",
-  "wnfs-bench": "0.1.16"
+  "wnfs-bench": "0.1.16",
+  "wnfs-common": "0.1.16",
+  "wnfs-hamt": "0.1.16",
+  "wnfs-namefilter": "0.1.16",
```
matheus23, line 6: Ah - I wish I said this sooner so I could've spared you some work: We probably don't need to extract out the namefilter, because (1) it's just a small wrapper around what should eventually be a dependency on the [deterministic-bloom](https://github.com/wnfs-wg/deterministic-bloom) package, and (2) we'll remove the namefilter code once we implement name accumulators, which will happen in the next 2-3 weeks (I'm working on the spec ATM).

appcypher: I'm aware of this. It wasn't hard to move out at all. My thinking was that it no longer needs to live in `wnfs` core just like the `deterministic-bloom` package.
So the transition to another dependnency should be easy.

pinkforest (approved):
Just a small nit but since these are mostly just renames / moving around things I think we should get this in as it's indeed blocking things including what I was going to propose :)

matheus23:
`cargo doc` generates two warnings for me:
```rs
warning: unresolved link to `crate::private::Node`
  --> wnfs-hamt/src/hamt.rs:23:82
   |
23 | ...de`](crate::private::Node).
   |         ^^^^^^^^^^^^^^^^^^^^ no item named `private` in module `wnfs_hamt`
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: unresolved link to `crate::HashOutput`
  --> wnfs-hamt/src/hash.rs:13:19
   |
13 | /// [HashOutput]: crate::HashOutput
   |                   ^^^^^^^^^^^^^^^^^ no item named `HashOutput` in module `wnfs_hamt`

warning: `wnfs-hamt` (lib doc) generated 2 warnings
```

wnfs-common/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+<!-- Todo: Summary here -->
+
+## Usage
+
+<!-- Todo: Usage description here -->
+
+```rust
+todo!()
+```
```
matheus23, line 42: Intended leftover TODOs, or should they be attended to this PR?

wnfs-common/src/blockstore.rs
```diff
@@ -23,19 +24,26 @@ pub trait BlockStore {
...
     }
 
-    async fn get_deserializable<V: DeserializeOwned>(&self, cid: &Cid) -> Result<V> {
+    async fn get_deserializable<'a, V: DeserializeOwned>(&'a self, cid: &Cid) -> Result<V> {
         let bytes = self.get_block(cid).await?;
-        let ipld = dagcbor::decode(bytes.as_ref())?;
+        let ipld = Ipld::decode(DagCborCodec, &mut Cursor::new(bytes.as_ref()))?;
         Ok(ipld_serde::from_ipld::<V>(ipld)?)
     }
 }
```
matheus23, line 49: This is undoing some changes that I think were good, presumably due to a merge conflict. Shouldn't we change them back?

wnfs-common/src/lib.rs
```diff
@@ -1,10 +1,13 @@
+//! TODO(appcypher): Needs documentation.
```
matheus23, line 1: When opening up the docs, this is what I see at the top.

Intended to leave, or do this PR?

wnfs-hamt/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+<!-- TODO(appcypher): Needs documentation -->
+
+## Usage
+
+<!-- TODO(appcypher): Needs documentation -->
+
+```rust
+todo!()
+```
```
matheus23, line 42: Some TODOs

wnfs-common/src/utils.rs
```diff
@@ -0,0 +1,128 @@
...
+/// ```
+/// use wnfs_common::utils;
+///
+/// let digest = utils::make_digest(&[0xff, 0x22]);
+///
+/// assert_eq!(digest.len(), 32);
+/// ```
+///
+/// [HashOutput]: crate::HashOutput
+pub fn make_digest(bytes: &[u8]) -> HashOutput {
```
matheus23, line 124: Perhaps we should make this `#[cfg(test)]`?
Or rename from `make_digest`, because this may get mistaken with an actual hashing function.
I know this code wasn't written in this PR. I'm not sure, but I *think* this PR made the function public, though (and I realize that's necessary when splitting into crates :) )

wnfs-namefilter/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+<!-- TODO(appcypher): Needs documentation. -->
+
+## Usage
+
+<!-- TODO(appcypher): Needs documentation -->
+
+```rust
+todo!()
+```
```
matheus23, line 42: TODOs! :)

wnfs-namefilter/src/lib.rs
```diff
@@ -1,5 +1,6 @@
+//! TODO(appcypher): Needs documentation.
```
matheus23, line 1: Same Q: Should this happen this PR?
I can see "no" being fine, given that we may swap this out for name accumulators soon (TM)

wnfs/src/private/node/mod.rs
```diff
@@ -0,0 +1,8 @@
+mod header;
+mod keys;
```
matheus23, line 2: Thanks for splitting these out! :) Love this change.

wnfs/src/public/link.rs
```diff
@@ -12,27 +13,69 @@ use crate::Link;
 /// A link to another node in the WNFS public file system. It can be held as a simple serialised CID or as a reference to the node itself.
 ///
 /// The public file system is a DAG so we don't have to worry bout cyclic references.
-pub type PublicLink = Link<PublicNode>;
+#[derive(Debug, Clone, PartialEq)]
+pub struct PublicLink(Link<PublicNode>);
```
matheus23, line 17: I'm in favor of going with `Link<PublicNode>` in use sites instead of `PublicLink`.
This makes code just ever so slightly shorter, but I'm not sure whether it's worth it.
Not a deal-breaker at all! Just throwing it out there. You decide :)

wnfs/src/lib.rs
```diff
@@ -8,26 +8,27 @@
...
 }
+pub mod common {
+    pub use wnfs_common::*;
+}
+pub mod hamt {
+    pub use wnfs_hamt::*;
+}
+pub mod namefilter {
+    pub use wnfs_namefilter::*;
+}
```
matheus23, line 34: Just so I understand this correctly:
The reasoning for re-exporting generally is that users of `wnfs` shouldn't have to import another library if they want to make use of its public API, right?
So, for `common` and `namefilter` that makes sense to me. But do `hamt` details actually appear in the public API or is everything hidden behind the `PrivateForest`?
Can we just keep the HAMT detail hidden & not reexport?

matheus23 (approved):
Looking good! :)

:+1: from me after some nits, but most things are just prompts to decide about whether to act on or not.

wnfs-common/README.md
pinkforest: imo we can add that on another  PR

pinkforest: Also I think I would recommend having examples directory and treating them as artifacts
docs.rs has learned to harves them and inject into relevant places
e.g. we could remove it maybe

wnfs-common/src/blockstore.rs
appcypher: I suspect some of the rebases I did undid some of your changes. Tried to find all of them.

wnfs-common/src/lib.rs
pinkforest:
```suggestion
//! WNFS Utils
//------------------------------------------------------------------------
// Documentation:
//------------------------------------------------------------------------
#![deny(missing_docs)]
#![doc = include_str!("../README.md")]
```

wnfs/src/lib.rs
appcypher: There is the `Hasher` trait

wnfs-hamt/README.md
pinkforest:
```suggestion
```

imo cover these in examples/ :)

wnfs-namefilter/src/lib.rs
pinkforest:
```suggestion
//! WNFS Namefilter
//------------------------------------------------------------------------
// Documentation:
//------------------------------------------------------------------------
#![deny(missing_docs)]
#![doc = include_str!("../README.md")]
```

wnfs/src/lib.rs
matheus23: Right.

What if we put that into wnfs-common?

wnfs/src/public/link.rs
appcypher: We won't be able to specialize, if we need to. Like with `with_dir`. Probably not great examples

PR #197, opened 2023-03-01
fix: `find_latest_share_counter` finds the last share count

### LINKED ISSUE #196: `find_share` should return the last count
# Summary

## Problem

`find_share` returns the first count.

### Impact

Can't figure out the counter needed to make a new share.

## Solution

There should be a function that returns the counter needed to make a new share.

matheus23:
## Summary

Fixes #196 

The idea is to provide a function that finds the maximum share count for available shares. This allows callers to check all share counts they might have missed in between.

## Test plan (required)

I added another test that checks after a bunch of shares that the maximum share counter is returned.

### DIFF
--- a/wnfs/src/private/share.rs
+++ b/wnfs/src/private/share.rs
@@ -326,9 +326,7 @@ pub mod recipient {
         sharer_forest: &PrivateForest,
         store: &impl BlockStore,
     ) -> Result<Option<u64>> {
-        for i in 0..limit {
-            let share_count = share_count_start + i;
-
+        for share_count in share_count_start..share_count_start + limit {
             let share_label =
                 sharer::create_share_label(share_count, sharer_root_did, recipient_exchange_key);
 
@@ -336,8 +334,8 @@ pub mod recipient {
                 .has(&Sha3_256::hash(&share_label), store)
                 .await?
             {
-                if i == 0 {
-                    // There don't seem to be any shares yet
+                if share_count == share_count_start {
+                    // There don't seem to be any shares
                     return Ok(None);
                 } else {
                     // We've hit the first unpopulated label,
--- a/wnfs/src/private/share.rs
+++ b/wnfs/src/private/share.rs
@@ -317,9 +317,9 @@ pub mod recipient {
     use wnfs_hamt::Hasher;
     use wnfs_namefilter::Namefilter;
 
-    /// Checks if a share count is available.
-    pub async fn find_share(
-        share_count: u64,
+    /// Seeks to the latest share counter that is populated.
+    pub async fn find_latest_share_counter(
+        share_count_start: u64,
         limit: u64,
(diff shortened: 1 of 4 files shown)


matheus23:
Okay, fixed the wasm. CI is green now :)

appcypher (approved):
This looks great to me. I only have minor suggestions that don't really affect the code.
Thanks for fixing this!

wnfs/src/private/share.rs
```diff
@@ -317,31 +317,38 @@ pub mod recipient {
...
         sharer_forest: &PrivateForest,
         store: &impl BlockStore,
     ) -> Result<Option<u64>> {
         for i in 0..limit {
-            let share_label = sharer::create_share_label(
-                share_count + i,
-                sharer_root_did,
-                recipient_exchange_key,
-            );
+            let share_count = share_count_start + i;
```
appcypher, line 330: Nit: Not that it matters much

```rs
for share_count in share_count_start..(share_count_start + limit) {
    // ...
```

wnfs/src/private/share.rs
```diff
@@ -317,31 +317,38 @@ pub mod recipient {
...
             {
-                return Ok(Some(share_count + i));
+                if i == 0 {
+                    // There don't seem to be any shares yet
+                    return Ok(None);
+                } else {
+                    // We've hit the first unpopulated label,
+                    // so the last one was the last valid share counter.
+                    return Ok(Some(share_count - 1));
+                }
```
appcypher, line 346: Nit: Cosmetic suggestion :)

```rs
return Ok(...)
```

wnfs/src/private/share.rs
matheus23: I like it :) Will update :+1:

wnfs/src/private/share.rs
matheus23: Hm, I don't understand?

matheus23: Are you referring to the last line? clippy actually complains if I put `return Ok(Some(...));` into the last line instead of `Ok(Some(...))`.

appcypher: Sorry. I mean 

```rs
return Ok(if i == 0 {
    // There don't seem to be any shares yet
    None
} else {
    // We've hit the first unpopulated label,
    // so the last one was the last valid share counter.
    Some(share_count - 1)
})
```

You can ignore if it doesn't appeal to your senses.

matheus23: Oh. I actually had it like this before, but ended up changing it.
Especially with the above change it's going to be a bigger check than `i == 0`, so perhaps more readable if it's not an expression :D
I'll keep it this way.

appcypher: That makes sense

PR #201, opened 2023-03-09
refactor(api): refactor out `PublicOpResult`

appcypher:
## Summary

Removed the need for `PublicOpResult` type for capturing results of operating on an immutable `PublicDirectory`

## Test plan (required)

```bash 
scripts/rs-wnfs.sh test
```

## Closing issues

Fixes https://github.com/wnfs-wg/rs-wnfs/issues/185


wnfs/examples/public.rs
```diff
@@ -15,16 +12,16 @@ async fn main() {
...
         .mkdir(&["pictures".into(), "cats".into()], Utc::now(), &store)
         .await
         .unwrap();
 
     // Add a file to /pictures/dogs directory.
-    let PublicOpResult { root_dir, .. } = root_dir
+    root_dir
         .write(
             &["pictures".into(), "dogs".into(), "billie.jpeg".into()],
             Cid::default(),
```
matheus23, line 27: :metal:

wnfs/src/public/file.rs
```diff
@@ -142,6 +125,24 @@ impl PublicFile {
...
+        };
+
+        let cloned = Rc::make_mut(self);
+        cloned.persisted_as = OnceCell::new();
+        cloned.previous = [previous_cid].into_iter().collect();
+
+        cloned
+    }
+}
+
```
matheus23, line 145: What is the reasoning behind introducing a trait? (I'm generally very careful about doing that, which is why I'm asking)
I don't see any function that is abstract over an `impl PrepareMut` or `P: PrepareMut` or similar.

wnfs-wasm/tests/public.spec.ts
```diff
@@ -178,9 +178,9 @@ test.describe("PublicDirectory", () => {
         store
       );
 
-      var { result: imagesContent } = await rootDir.ls(["images"], store);
+      const imagesContent = await rootDir.ls(["images"], store);
 
-      var { result: picturesContent } = await rootDir.ls(["pictures"], store);
+      const picturesContent = await rootDir.ls(["pictures"], store);
```
matheus23, line 184: Cool :+1: I'm onboard with this change for read-only functions on the wasm-wnfs side :)

matheus23 (approved):
It's beautiful

starry eyes GIF

Just a small nit about `PrepareMut`, but good to go either way :)

---

Same thing on the private side next? :pleading_face:

wnfs/src/public/file.rs
appcypher: It is only there for reuse across the public and private side.

matheus23: Sorry, allow me to dig into this a little more:
Even if we had an `impl PrepareMut for PrivateDirectory` and `impl PrepareMut for PrivateFile`, etc.
we still wouldn't actually have any function that is abstract over "something that is `impl PrepareMut`". Everywhere that we would use `x.prepare_mut()`, we'd actually know the concrete type of `x`.
In these cases, the code would also work if we didn't use a trait for `.prepare_mut()`.

appcypher: It is just used to describe a common interface. It can be rid of; not that it matters too much.

matheus23: Okay. Yeah so we're somewhat on the same page. I just tend to avoid traits as much as possible. (This is me coming from haskell, where people have been "overusing" type classes a little too much, and which resulted in some people taking on [strong stances on using type classes](https://stackoverflow.com/a/17100055). Of course rust != haskell, but I think some insights transfer.)

PR #208, opened 2023-03-20
chore: Use cargo-deny & fix time CVE

matheus23:
Basically just following what we've been doing in rust-template (https://github.com/fission-codes/rust-template/pull/26) and porting this over to here.


pinkforest (approved):
lgtm 💜

PR #211, opened 2023-03-22
refactor(api): refactor out PrivateOpResult 

appcypher:
## Summary

Removed the need for `PrivateOpResult` type for capturing results of operating on an immutable `PrivateDirectory`

## Test plan (required)

```bash 
scripts/rs-wnfs.sh test
```

### DIFF
--- a/wnfs-common/src/link.rs
+++ b/wnfs-common/src/link.rs
@@ -81,7 +81,7 @@ impl<T: RemembersCid> Link<T> {
     /// Gets mut value stored in link. It attempts to get it from the store if it is not present in link.
     pub async fn resolve_value_mut(&mut self, store: &(impl BlockStore + ?Sized)) -> Result<&mut T>
     where
-        T: DeserializeOwned, // + Clone,
+        T: DeserializeOwned,
     {
         match self {
             Self::Encoded { cid, value_cache } => {
@@ -125,14 +125,11 @@ impl<T: RemembersCid> Link<T> {
         T: DeserializeOwned,
     {
         match self {
-            Self::Encoded {
-                ref cid,
-                value_cache,
-            } => match value_cache.into_inner() {
+            Self::Encoded { cid, value_cache } => match value_cache.into_inner() {
                 Some(cached) => Ok(cached),
                 None => {
-                    let value: T = store.get_deserializable(cid).await?;
-                    value.persisted_as().get_or_init(async { *cid }).await;
+                    let value: T = store.get_deserializable(&cid).await?;
+                    value.persisted_as().get_or_init(async { cid }).await;
                     Ok(value)
                 }
             },
--- a/wnfs/src/private/directory.rs
+++ b/wnfs/src/private/directory.rs
@@ -1725,8 +1725,7 @@ mod tests {
     async fn search_latest_finds_the_most_recent() {
         let store = &mut MemoryBlockStore::default();
         let forest = &mut Rc::new(PrivateForest::new());
-        let rng = &mut TestRng::deterministic_rng(RngAlgorithm::ChaCha);
-        // let rng = &mut rand::thread_rng();
+        let rng = &mut rand::thread_rng();
         let root_dir = &mut Rc::new(PrivateDirectory::new(
             Namefilter::default(),
             Utc::now(),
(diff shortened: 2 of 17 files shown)


wnfs-common/src/link.rs
```diff
@@ -125,14 +125,11 @@ impl<T: RemembersCid> Link<T> {
...
-                ref cid,
-                value_cache,
-            } => match value_cache.into_inner() {
+            Self::Encoded { cid, value_cache } => match value_cache.into_inner() {
                 Some(cached) => Ok(cached),
                 None => {
-                    let value: T = store.get_deserializable(cid).await?;
-                    value.persisted_as().get_or_init(async { *cid }).await;
+                    let value: T = store.get_deserializable(&cid).await?;
+                    value.persisted_as().get_or_init(async { cid }).await;
```
matheus23, line 132: We probably need the same fix here.

If we're in `Self::Encoded`, then resolve the `value_cache` (or load the cid), and then set `*self = Self::Decoded`, before returning `&mut T`.

wnfs/src/private/file.rs
```diff
@@ -561,28 +561,24 @@ impl PrivateFile {
     /// This doesn't have any effect if the current state hasn't been `.store()`ed yet.
     /// Otherwise, it clones itself, stores its current CID in the previous links and
     /// advances its ratchet.
-    pub(crate) fn prepare_next_revision(self: Rc<Self>) -> Result<Self> {
+    pub(crate) fn prepare_next_revision<'a>(self: &'a mut Rc<Self>) -> Result<&'a mut Self> {
+        println!("persisted_as: {:?}", self.content.persisted_as.get());
```
matheus23, line 565: :mute:

wnfs/src/private/directory.rs
```diff
@@ -540,30 +591,35 @@ impl PrivateDirectory {
...
         match self
-            .get_path_nodes(path, search_latest, forest, store)
+            .get_leaf_dir(path, search_latest, forest, store)
             .await?
         {
-            PathNodesResult::Complete(node_path) => {
-                match node_path
-                    .tail
+            SearchResult::Found(dir) => {
+                println!(
```
matheus23, line 607: :mute:

wnfs/src/private/directory.rs
```diff
@@ -540,30 +591,35 @@ impl PrivateDirectory {
...
+                    search_latest, dir.header.ratchet
+                );
+                match dir
                     .lookup_node(filename, search_latest, forest, store)
                     .await?
                 {
                     Some(PrivateNode::File(file)) => {
-                        let result = file.get_content(forest, store).await?;
-                        Ok(PrivateOpResult { root_dir, result })
+                        println!(
```
matheus23, line 616: :mute:

wnfs/src/private/directory.rs
```diff
@@ -540,30 +591,35 @@ impl PrivateDirectory {
     /// }
     /// ```
     pub async fn read(
-        self: Rc<Self>,
+        self: &Rc<Self>,
```
matheus23, line 594: Huh, what's the reason for `&`ing this? Just consistency?
I think a `mut Rc<PrivateDirectory>` or `&mut Rc<PrivateDirectory>` should deref to `Rc<Self>` just fine, no?

wnfs/src/private/directory.rs
```diff
@@ -622,31 +677,31 @@ impl PrivateDirectory {
...
         store: &mut impl BlockStore,
         rng: &mut impl RngCore,
-    ) -> Result<PrivateOpResult<()>> {
-        let (directory_path, filename) = split_last(path_segments)?;
-
-        // This will create directories if they don't exist yet
-        let mut directory_path_nodes = self
-            .get_or_create_path_nodes(directory_path, search_latest, time, forest, store, rng)
+    ) -> Result<()> {
+        println!();
```
matheus23, line 689: :mute:

wnfs/src/private/directory.rs
```diff
@@ -657,100 +712,28 @@ impl PrivateDirectory {
...
+                    dir.header.bare_name.clone(),
                     time,
                     content,
                     forest,
                     store,
                     rng,
                 )
-                .await?
+                .await?;
+                println!(
```
matheus23, line 727: :mute:

wnfs/src/private/directory.rs
matheus23: Oh yeah right. Not having the `&` would force the caller to `Rc::clone`!

wnfs-common/src/link.rs
appcypher: Can you suggest a fix for this?

matheus23:
```rs
pub async fn resolve_value_mut(&mut self, store: &(impl BlockStore + ?Sized)) -> Result<&mut T>
where
    T: DeserializeOwned,
{
    match self {
        Self::Encoded { cid, value_cache } => {
            let value = match value_cache.take() {
                Some(v) => v,
                None => {
                    let value: T = store.get_deserializable(cid).await?;
                    value.persisted_as().get_or_init(async { *cid }).await;
                    value
                }
            };

            *self = Self::Decoded { value };

            Ok(match self {
                Self::Encoded { .. } => unreachable!(),
                Self::Decoded { value } => value,
            })
        }
        Self::Decoded { value, .. } => Ok(value),
    }
}
```

appcypher: Yeah, you were probably referring to `resolve_value_mut` and not `resolve_owned_value`

matheus23 (approved):
Last puzzle piece

Last piece of the puzzle landed :raised_hands:

PR #218, opened 2023-03-26
feat: `open_file_mut` function for getting `&mut PrivateFile` references

### LINKED ISSUE #210: API for "open"ing a file path and getting a &mut PrivateFile
Something like

```rs
impl PrivateDirectory {
    // ...

    pub async fn open(
        self: &mut Rc<Self>,
        path_segments: &[String],
        search_latest: bool,
        time: DateTime<Utc>,
        forest: &mut Rc<PrivateForest>,
        store: &mut impl BlockStore,
        rng: &mut impl RngCore
    ) -> Result<&mut PrivateFile>;

}
```

If the file is missing, it would initialize an empty file and give a mut reference to that.
If the file already exists, it would copy it to the next revision and give a mut reference to that.

Enables creating and exposing functions like

```rs
impl PrivateFile {
    // ...

    /// Overwrites any file content currently set to the content read from the stream
    pub async fn write_streaming(
        &mut self,
        content: impl AsyncRead,
        forest: &mut Rc<PrivateForest>,
        store: &mut impl BlockStore,
        rng: &mut impl RngCore,
    );
    
    pub fn get_metadata_mut(&mut self) -> &mut Metadata;
}
```

So that the flow for e.g. writing to a file looks like this:

```rs
let mut root: Rc<PrivateDirectory> = // ... some root directory;
let file = root.open(&["Docs".into(), "Notes.md".into()], true, Utc::now(), forest, store, rng).await?;
file.write_streaming(File::open("/tmp/somefile.md")?, forest, store, rng).await?;
file.get_metadata_mut().put("written_by", Ipld::String("James"));
```

laudiacay:
## Summary

- [x] Adds open_file_mut so you can change the content of files and write them back

Explain the **motivation** for making this change. What existing problem does the pull request solve?
See #210- prep for integrating with Filecoin

## Test plan (required)

This doesn't pass doctests- I'm pretty sure without adding some interior mutability, we can't make the desired workflow in #210 work. open to being told I'm wrong. The doctest error is illustrative, and I gave more details in #210.

## Closing issues

Closes #210 

## After Merge

- [ ] Does this change require a release to be made? Is so please create and deploy the release
(none of the above)

### DIFF
--- a/wnfs/src/private/directory.rs
+++ b/wnfs/src/private/directory.rs
@@ -633,7 +633,7 @@ impl PrivateDirectory {
     /// async fn main() {
     ///    let store = &mut MemoryBlockStore::default();
     ///    let rng = &mut thread_rng();
-    ///    let mut forest : &mut Rc<PrivateForest> = &mut Rc::new(PrivateForest::new());
+    ///    let forest = &mut Rc::new(PrivateForest::new());
     ///    let root_dir = &mut Rc::new(PrivateDirectory::new(
     ///         Namefilter::default(),
     ///         Utc::now(),
@@ -646,27 +646,23 @@ impl PrivateDirectory {
     ///            true,
     ///            Utc::now(),
     ///            content.to_vec(),
-    ///            &mut forest,
+    ///            forest,
     ///            store,
     ///            rng
     ///        )
     ///        .await
     ///        .unwrap();
-    ///    let mut file_forest = &mut Rc::clone(forest);
-    ///    let mut file = {
-    ///        root_dir
-    ///            .open_file_mut(&["code".into(), "hello.py".into()], true, Utc::now(), forest, store, rng)
-    ///            .await
-    ///            .unwrap()
-    ///    };
+    ///    let mut file = root_dir
+    ///        .open_file_mut(&["code".into(), "hello.py".into()], true, Utc::now(), forest, store, rng)
+    ///        .await
+    ///        .unwrap();
     ///    file.set_content(
     ///        Utc::now(),
     ///        &b"print('hello world 2')"[..],
-    ///        file_forest,
+    ///        forest,
     ///        store,
     ///        rng,
-    ///    );
-    ///    let forest = &forest.merge(&file_forest, store).await.unwrap();
+    ///    ).await.unwrap();
     ///    let result = root_dir
     ///        .read(&["code".into(), "hello.py".into()], true, forest, store)
     ///        .await
@@ -679,14 +675,14 @@ impl PrivateDirectory {
         path_segments: &[String],
         search_latest: bool,
         time: DateTime<Utc>,
-        forest: &'a mut Rc<PrivateForest>,
-        store: &'a mut impl BlockStore,
+        forest: &mut Rc<PrivateForest>,
+        store: &mut impl BlockStore,
         rng: &mut impl RngCore,
     ) -> Result<&'a mut PrivateFile> {
         let (path, filename) = crate::utils::split_last(path_segments)?;
-        let SearchResult::Found(dir) = self.get_leaf_dir_mut(path, search_latest, forest, store).await? else {
-            bail!(FsError::NotFound);
-        };
+        let dir = self
+            .get_or_create_leaf_dir_mut(path, time, search_latest, forest, store, rng)
+            .await?;
 
         if !dir.content.entries.contains_key(filename.as_str()) {
             let parent_bare_name = dir.header.bare_name.clone();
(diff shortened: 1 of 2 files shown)


wnfs/src/private/directory.rs
```diff
@@ -616,6 +616,94 @@ impl PrivateDirectory {
...
+    ///        &b"print('hello world 2')"[..],
+    ///        forest,
+    ///        store,
+    ///        rng,
+    ///    );
+    ///    let result = root_dir
+    ///        .read(&["code".into(), "hello.py".into()], true, forest, store)
+    ///        .await
+    ///        .unwrap();
+    ///    assert_eq!(&result, b"print('hello world 2')");
```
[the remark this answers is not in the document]
laudiacay: i think an RwLock-based one still won't satisfy the borrow checker because there are multiple consecutive functions calling wanting `store: &'a mut impl BlockStore` as an argument.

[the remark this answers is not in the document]
matheus23: > i think an RwLock-based one still won't satisfy the borrow checker because there are multiple consecutive functions calling wanting `store: &'a mut impl BlockStore` as an argument.

Yeah that's true, but a `RwLock`-based block store should enable a cheap `.clone()` implementation. Then it should be possible to provide two `&mut impl BlockStore`s.

That said, I'm currently wrapping my head around "what if it's just `impl BlockStore`" and whether it's implemented on `&mut MyBlockStore` or `&MyBlockStore` should be (literally) an implementation detail :thinking: 

We may be able to use rust better here.

wnfs/src/private/directory.rs
```diff
@@ -616,6 +616,96 @@ impl PrivateDirectory {
...
+    /// }
+    /// ```
+    pub async fn open_file_mut<'a>(
+        self: &'a mut Rc<Self>,
+        path_segments: &[String],
+        search_latest: bool,
+        time: DateTime<Utc>,
+        forest: &'a mut Rc<PrivateForest>,
+        store: &'a mut impl BlockStore,
```
matheus23, line 683: I took another look at this today, I think this works :)

```suggestion
        forest: &mut Rc<PrivateForest>,
        store: &mut impl BlockStore,
```

No need for `forest.merge` then and we can also re-use the same BlockStore :+1:

matheus23:
Not sure if the push was intentional or not, but I'd like to separate the `open_file_mut` work from the BlockStore work, they should be separate PRs :v:

Also, the `open_file_mut` stuff is really close. It just needs:
- My latest suggestion from above (using `&mut` instead of `&'a mut`)
- An `.await.unwrap()` in the doctest on the result of the `file.set_content` call
- Some cleanup in the doctest (e.g. no unnecessary type sigs or scopes)

wnfs/src/private/directory.rs
```diff
@@ -616,6 +616,94 @@ impl PrivateDirectory {
...
+    ///            rng
+    ///        )
+    ///        .await
+    ///        .unwrap();
+    ///    let mut file = {
+    ///        root_dir
+    ///            .open_file_mut(&["code".into(), "hello.py".into()], true, Utc::now(), forest, store, rng)
+    ///            .await
+    ///            .unwrap()
+    ///    };
```
matheus23, line 660: I think you can remove the extra scope here.
```suggestion
    ///    let mut file = root_dir
    ///        .open_file_mut(&["code".into(), "hello.py".into()], true, Utc::now(), forest, store, rng)
    ///        .await
    ///        .unwrap();
```

wnfs/src/private/directory.rs
```diff
@@ -616,6 +616,94 @@ impl PrivateDirectory {
...
+    /// use wnfs::{
+    ///    private::{PrivateForest, PrivateRef, PrivateDirectory},
+    ///    common::{BlockStore, MemoryBlockStore},
+    ///    namefilter::Namefilter,
+    /// };
+    /// #[async_std::main]
+    /// async fn main() {
+    ///    let store = &mut MemoryBlockStore::default();
+    ///    let rng = &mut thread_rng();
+    ///    let mut forest = &mut Rc::new(PrivateForest::new());
```
matheus23, line 636: This doesn't need to be `mut`, it's just a constant reference to something that's mutable:
```suggestion
    ///    let forest = &mut Rc::new(PrivateForest::new());
```

matheus23 (approved):
Just a couple of nits. Should be GTG after addressing them :)

Again, thanks for this :pray:

matheus23:
Great. Feel free to push the merge button once CI went through, you should have rights :)

PR #223, opened 2023-03-30
chore: Write up an example of how to use tiered blockstores

matheus23:
An example explaining how you can separate out file data from hierarchical data into different blockstores.

There's a bunch of ways of how to slice this problem.
You could technically also separate out the data into two `PrivateForest`s. This gives you two separate root CIDs to pin, if that's something that a subsystem understands.
This example is not that though. It simply bulids a single PrivateForest, and stores some of the blocks that its leaves point to in a different blockstore than the rest.
However, rs-wnfs is actually really flexible. Both versions are absolutely doable.

### DIFF
--- a/wnfs/examples/tiered_blockstores.rs
+++ b/wnfs/examples/tiered_blockstores.rs
@@ -1,0 +1,116 @@
+//! This example shows how to separate file system hierarchy data from file metadata at write-time,
+//! storing data about hierarchy in a block store that is kept 'hot' so reads can be made fast,
+//! and storing actual file content in a block store that is meant to be 'cold' as in, reads only
+//! work with high latency.
+
+use anyhow::Result;
+use async_trait::async_trait;
+use chrono::Utc;
+use libipld::{Cid, IpldCodec};
+use rand::thread_rng;
+use std::{borrow::Cow, rc::Rc};
+use wnfs::private::{PrivateDirectory, PrivateForest, PrivateNode};
+use wnfs_common::{BlockStore, MemoryBlockStore};
+use wnfs_namefilter::Namefilter;
+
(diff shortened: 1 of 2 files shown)


appcypher (approved):
This looks good to me! :tada:

wnfs/examples/tiered_blockstores.rs
```diff
@@ -0,0 +1,116 @@
...
+    );
+
+    let directory = PrivateNode::load(&private_ref, &forest, &hot_store)
+        .await
+        .unwrap()
+        .as_dir()
+        .unwrap();
+
+    // Reading the file's data will fail when only provided the hot store:
+    // assert!(directory.read(&file_path, true, &forest, &hot_store).await.is_err()); // currently panics for mysterious reasons?
```
matheus23, line 78: @appcypher can you help me understand why, when I uncomment this line, this causes a panic & stacktrace?
What I intended was simply to assert that the result of `.await` (without any `.unwrap()`) is an `Err`. But for some reason that doesn't work :thinking:

matheus23: Well, I *know* that there's an `Err`, and I *know* that the error is basically that a block is requested from `hot_store` which doesn't exist. But I *expect* that, and I just want to make sure with the assert that this is the case & document that.

matheus23: Ok, @appcypher and I figured this one out on a call. I submitted an issue: https://github.com/wnfs-wg/rs-wnfs/issues/228

matheus23: The issue is now fixed. Thanks @appcypher :)

PR #229, opened 2023-04-05
chore(api): mutable API preliminary work

appcypher:
## Summary

Initial work on mutable/ root tree API. There are still some work that needs to be done to complete the API (public/private root merge, blockstore ref only, etc.) but this is as far as we can go right now.

## Test plan (required)

```bash 
scripts/rs-wnfs.sh test
```

### DIFF
--- a/wnfs/src/root_tree.rs
+++ b/wnfs/src/root_tree.rs
@@ -26,8 +26,8 @@ use wnfs_common::{BlockStore, Metadata};
 //--------------------------------------------------------------------------------------------------
 
 #[derive(Debug)]
-pub struct RootTree<B: BlockStore, R: RngCore> {
-    pub store: B, // TODO(appcypher): Making put_block &self will remove the need for *RootMut types
+pub struct RootTree<'a, B: BlockStore, R: RngCore> {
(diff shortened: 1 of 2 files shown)


appcypher:
The unused lint errors can be ignored since the API is not public yet.

wnfs/src/root_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+use wnfs_common::MemoryBlockStore;
+use wnfs_common::{BlockStore, Metadata};
+
+//--------------------------------------------------------------------------------------------------
+// Types
+//--------------------------------------------------------------------------------------------------
+
+#[derive(Debug)]
+pub struct RootTree<B: BlockStore, R: RngCore> {
+    pub store: B, // TODO(appcypher): Making put_block &self will remove the need for *RootMut types
```
matheus23, line 30: Should we change it to `&B` now?

matheus23 (approved):

wnfs/src/root_tree.rs
appcypher: You are right

PR #237, opened 2023-04-18
feat: Add `PrivateDirectory::entires`, `PrivateFile::read_at` and make `PrivateFile::get_content_size_upper_bound` public

Frando:
This PR contains changes I needed to made while hacking on [wnfs-fuse](https://github.com/Frando/wnfs-fuse) at IPFSThing.

- Public getter method on `PrivateDirectory` to get the entries. `ls` exists, but loads all the metadata, which is expensive if not needed
- Make `PrivateFile::get_content_size_upper_bound` public (to have something to report as filesize - even if inaccurate). It might be a good idea to add the actual filesize to the metadata instead.
- Add a new method `PrivateFile::read_at` to read `size` bytes from `offset`. This makes use of `stream_content` internally and does the block and offset calculations

### DIFF
--- a/wnfs/src/private/directory.rs
+++ b/wnfs/src/private/directory.rs
@@ -1002,6 +1002,10 @@ impl PrivateDirectory {
         }
     }
 
+    pub fn entries<'a>(self: &'a Rc<Self>) -> impl Iterator<Item = &'a String> {
+        self.content.entries.iter().map(|x| x.0)
+    }
+
     /// Removes a file or directory from the directory.
     ///
     /// # Examples
--- a/wnfs/src/private/directory.rs
+++ b/wnfs/src/private/directory.rs
@@ -1002,7 +1002,11 @@ impl PrivateDirectory {
         }
     }
 
-    pub fn entries<'a>(self: &'a Rc<Self>) -> impl Iterator<Item = &'a String> {
+    /// Get the names of directory's immediate children.
+    ///
+    /// Other than [PrivateDirectory::ls] this returns only the names, without loading the
+    /// metadata for each node from the store.
+    pub fn get_entries<'a>(self: &'a Rc<Self>) -> impl Iterator<Item = &'a String> {
         self.content.entries.iter().map(|x| x.0)
     }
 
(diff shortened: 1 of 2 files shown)


appcypher:
Amazing work on `wnfs-fuse` and thanks for the contribution.
This looks great to me. Just minor nits.

wnfs/src/private/directory.rs
```diff
@@ -1002,6 +1002,10 @@ impl PrivateDirectory {
         }
     }
 
+    pub fn entries<'a>(self: &'a Rc<Self>) -> impl Iterator<Item = &'a String> {
+        self.content.entries.iter().map(|x| x.0)
+    }
```
appcypher, line 1007: Can you add a tiny doc to the function? 
Nit: Can you also rename the function `get_entries` just to be consistent with the rest of the code?

wnfs/src/private/file.rs
```diff
@@ -504,7 +549,7 @@ impl PrivateFile {
     }
 
     /// Gets the upper bound of a file content size.
-    pub(crate) fn get_content_size_upper_bound(&self) -> usize {
+    pub fn get_content_size_upper_bound(&self) -> usize {
```
appcypher, line 552: 👍🏾

wnfs/src/private/directory.rs
Frando: :+1:

Frando:
I also added a test for `PrivateFile::read_at`. Don't have much experience with proptests, but copied one of the other tests and adapted it.

wnfs/src/private/file.rs
```diff
@@ -342,6 +342,51 @@ impl PrivateFile {
...
+                    };
+                    bytes.extend_from_slice(&chunk[from..to]);
+                    if index == last_block {
+                        break;
+                    }
+                }
+                // If a block doesn't exist, break and return what we read until here.
+                Err(_err) => {
+                    break;
+                }
```
matheus23, line 384: Can you make this just error out?
We shouldn't return any bytes if the content is corrupted. Usually we'd do `InvalidSerialization` or w/e it was called, but adding a custom error type may not be bad either.

matheus23 (approved):
One critical change about erroring out required, but LGTM when that's handled :) :rocket:

wnfs/src/private/file.rs
Frando: Yup, makes sense.

Issue #228, opened 2023-04-03
Fix erroneous `.unwrap()`


matheus23:
# Summary

An `.unwrap()` in the `PrivateFile` implementation can panic, if the blockstore doesn't contain the blocks to read the file.

## Solution

We should just propagate out the result in this function:
https://github.com/wnfs-wg/rs-wnfs/blob/d77d3356953275e5bbd2645225c658b434a9ada3/wnfs/src/private/file.rs#L387-L400

There's `TryStreamExt`. It may have some [functions to do this](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_for_each).

PR #252, opened 2023-05-03
fix: propagate missing chunk error

appcypher:
## Summary

As it is, the file content streaming implementation can panic if there is a missing chunk.
We should avoid unwrap as much as possible in the code and allow user to handle errors. This PR fixes that with the `try_stream_ext`

## Test plan (required)

```bash 
scripts/rs-wnfs.sh test
```

## Closing issues

#228


matheus23 (approved):
Aw nice. The negative test is
chef kiss

Just needs newer clippy to be made happy.

Issue #222, opened 2023-03-29
getNode fails when multiple segments of the path do not exist


icidasset:
Calling `getNode([ "top-dir", "nested-dir", "file" ])` will result in an error instead of returning `null`. Where "top-dir" exists, but "nested-dir" and "file" do not.

Given error:
```
Error: Cannot get node: Invalid WNFS path
```

appcypher:
How do you suggest we go about it? 
You could use a workaround by handling the error or we can introduce a variant of getNode that returns null with invalid paths.

matheus23:
Since `getNode` already returns a `Result<Option<...>>`, and that's `Ok(None)` if the last part of the path is missing, it should simply also return `Ok(None)` when any other path segment is missing.

PR #253, opened 2023-05-03
fix!: get_node should return null on missing path

appcypher:
## Summary

rs-wnfs `get_node` fns in `PublicDirectory` and `PrivateDirectory` currently do not replicate the same behavior in webnative when a non-existent path segment is specified. It returns an`Err(...)` instead of a `Ok(None)` which translates to `null` on the js side.

This PR solves that although there is the little inconvenience that wasm-bindgen translates `Ok(None)` to `undefined` on the js side.

## Test plan (required)

```bash 
scripts/rs-wnfs.sh test
```

## Closing issues

#222

### DIFF
--- a/wnfs-wasm/src/fs/private/directory.rs
+++ b/wnfs-wasm/src/fs/private/directory.rs
@@ -166,11 +166,7 @@ impl PrivateDirectory {
                 .await
                 .map_err(error("Cannot get node"))?;
 
-            Ok(utils::create_private_op_result(
-                directory,
-                forest,
-                result.map(PrivateNode),
-            )?)
+            Ok(value!(result.map(PrivateNode)))
         }))
     }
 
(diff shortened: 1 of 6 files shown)


wnfs-wasm/src/fs/private/directory.rs
```diff
@@ -166,11 +166,7 @@ impl PrivateDirectory {
                 .await
                 .map_err(error("Cannot get node"))?;
 
-            Ok(utils::create_private_op_result(
-                directory,
-                forest,
-                result.map(PrivateNode),
-            )?)
```
matheus23, line 173: :+1: Let's make sure that we capture this change somewhere in the CHANGELOG, so that once @icidasset or someone upgrades to the next rs-wnfs version they're not surprised why it crashes again :sweat_smile:

matheus23 (approved):
LGTM! :) Just need to make newer clippy happy.

PR #285, opened 2023-06-20
feat: make AccessKey the key entrypoint for public apis

### LINKED ISSUE #249: Unique canonical "entrypoint" API for private WNFS
Today we have
- `PrivateRef`
- `RevisionRef`
- `SnapshotKey` / `TemporalKey`
- `SharePayload`
- `SnapshotSharePointer` / `TemporalSharePointer`

We should probably just rename one of these (likely `SharePayload` and related) and make all our examples based around it.

Let's call it `AccessKey` for now (even though it contains the HAMT label & disambiguation CID as well) (totally up for renames).
`AccessKey` should be serde (`De`)`Serializable`, and there should be an easy way to go from `PrivateNode` to `AccessKey` and from forest + access key to node, possibly with shortcuts for directories/files.

The share functions in WNFS should make use of `AccessKey` (and by default serialize them as dag-cbor). In the future there will be a higher-level API that simply takes some asymmetric key stuff and does the right thing regarding secret management by using the share API "with itself".

The `PrivateRef` should not be exposed in the future. Instead everyone should use `AccessKey`.

### Seed-phrase-based secret management

APIs like `with_seed` in `RevisionRef` should be removed.

In the future the way to do this is create an *asymmetric key pair* for encryption (an exchange key) based on a seed phrase and then share the directory with yourself (i.e. storing the encrypted `AccessKey` in your `PrivateForest` at a predictable location). Then with the seed phrase you'll be able to re-derive the same asymmetric key pair & decrypt the corresponding "share payload" (`AccessKey`) in the `PrivateForest`.

appcypher:
This PR brings the multiple ways of generating key materials for accessing nodes in a filesystem under a single type. The `AccessKey` now becomes the only way of accessing the nodes in a file system. It is important to note that the Snapshot aspect of AccessKey is not at all functional right now.  

- [x] AccessKey implementation
- [x] Wasm binding for AccessKey

## Test plan (required)
-  Testing
    
    ```bash
    scripts/rs-wnfs test
    ```

## Closing issues

Fixes #249


matheus23:
Same with this PR, I'd prefer this to be marked as "feat" instead of "fix".
(Also I don't think the "(api)" means anything to release-please. IIUC, you can say e.g. "feat(wnfs-common)" or "feat(wnfs-wasm)" to indicate that it's only a change in a certain crate)

matheus23:
This needs conflicts resolved. Then I'd give it a look, we can merge and finally release :)

wnfs/examples/tiered_blockstores.rs
```diff
@@ -44,16 +44,20 @@ async fn main() -> Result<()> {
...
+    file.set_content(Utc::now(), &video[..], forest, &cold_store, rng)
         .await?;
 
     // When storing the hierarchy data blocks, we use the `hot_store`:
-    let private_ref = directory.store(forest, &mut hot_store, rng).await.unwrap();
+    let access_key = directory
+        .as_node()
+        .store(forest, &hot_store, rng)
+        .await
+        .unwrap();
```
matheus23, line 60: :+1:

wnfs/src/private/file.rs
```diff
@@ -666,41 +666,7 @@ impl PrivateFile {
...
-    ///     let node = PrivateNode::File(Rc::clone(&file));
-    ///
-    ///     assert_eq!(
-    ///         PrivateNode::load(&private_ref, forest, store).await.unwrap(),
-    ///         node
-    ///     );
-    /// }
-    /// ```
-    pub async fn store(
+    pub(crate) async fn store(
```
matheus23, line 669: :+1:

wnfs/src/private/forest.rs
```diff
@@ -74,9 +74,9 @@ impl PrivateForest {
     ///     ));
     ///
     ///     let node = PrivateNode::Dir(dir);
-    ///     let private_ref = node.store(forest, store, rng).await.unwrap();
+    ///     let access_key = node.store(forest, store, rng).await.unwrap();
     ///
-    ///     assert!(forest.has(&private_ref.saturated_name_hash, store).await.unwrap());
+    ///     assert!(forest.has(&access_key.get_label(), store).await.unwrap());
```
matheus23, line 79: :+1: :sparkles: 

Reads much better IMO :)

wnfs/src/private/forest.rs
```diff
@@ -140,24 +140,19 @@ impl PrivateForest {
...
-    /// Returns a stream of all private nodes that could be decrypted at given revision.
-    ///
-    /// The stream of results is ordered by CID.
-    ///
-    /// Each item in the resulting stream represents an instance of a concurrent write.
-    pub fn get_multivalue<'a>(
+    pub(crate) fn get_multivalue_with_revision_ref<'a>(
         &'a self,
-        revision: &'a RevisionRef,
+        revision_ref: RevisionRef,
```
matheus23, line 145: Yeah if we make this private, that may be an opportunity to rethink the revision ref again.

I think we can remove it, we practically pretty much only need it for this place in the API.
And internally, it's not too bad if instead of `RevisionRef` we need to pass `TemporalKey` and `HashOutput` as separate arguments.

I guess in the future getting multivalues should not be part of the API, instead (when we have conflict reconciliation) you'll only work with `AccessKey` and it'll merge it with the other multivalues in the same revision automatically.

Anyway, I don't know what the right move atm is. Perhaps we can pub(crate) the mutlivalue API completely?

In any case, I'm not even that sure, so that could happen in a separate PR.

wnfs/src/private/share.rs
```diff
@@ -2,18 +2,15 @@
 //! asynchronously when the other party may be offline using store-and-forward networks.
 //! The protocol uses asymmetric encryption, with RSA public keys, and is versioned to support multiple versions on the same file system.
 //! Public keys are widely distributed in a "exchange keys partition" and are grouped by device for easy selection by the sender.
-//! Payloads containing pointers to the private data are stored in the "Private Forest" and are labeled with a name filter that includes the sender's and recipient's information,
+//! Access keys containing pointers to the private data are stored in the "Private Forest" and are labeled with a name filter that includes the sender's and recipient's information,
```
matheus23, line 5:
```suggestion
//! Asymmetrically encrypted access keys containing pointers to the private data are stored in the "Private Forest" and are labeled with a name filter that includes the sender's and recipient's information,
```

matheus23 (approved):
:sparkles: 

I like these!

We learned that some things are useful on its own (SharePayload), rename them and sprinkle them along the API more.
We learned that you shouldn't create seeded keys and directories themselves, but instead seed an asymmetric keypair. Thus making a bunch of `RevisionRef` stuff obsolete.
We'll continue to learn more about exactly how we should best structure stuff.
I'm pretty confident we're continually improving :)

I'd love to have this merged, so I can do a final rebase of #247 and finally have a review & merge on that too.

wnfs/src/private/forest.rs
appcypher: Even if we do have conflict reconciliation, I think `get_multivalue` is a useful thing when you need to know what the conflicts are.

I agree RevisionRef feels unnecessary at this point. Let me see if I can get rid of it entirely.

appcypher: Decided to leave RevisionRef to a separate PR because it requires possibly deeper changes to how privateref is derived.

PR #257, opened 2023-05-05
Modified BlockStore trait

organizedgrime:
Here, many changes have been made, but all within the confines of `BlockStores` and how they operate. 

`BlockStore`s are no longer passed around mutably. Instead, an individual implementation of a given `BlockStore` will use `RefCell`s, `Lock`s, `Cow`s, etc. to achieve interior mutability in cases where it is required. This changes the signature and lifetimes for many functions, including the `put_block` and `get_block` functions required by the trait itself.

PR #286, opened 2023-06-21
feat: make changes to BlockStore trait based on feedback

### LINKED ISSUE #251: Feedback refactor: `BlockStore` restructure
Some feedback from dig:

- Adopting `bytes::Bytes` for `BlockStore` is not a terrible idea. It's even `no_std`.
- `IpldCodec` is pulling in a perhaps unnecessarily heavy dependency on `libipld`. What about making that simply a `u64`?
- `put_block` in practice is often taking `&self` instead of `&mut self`, since most people put concurrency primitives *inside* of their blockstore.

cc @organizedgrime @laudiacay because I know you were looking at blockstore stuff too.

appcypher:
This pull request addresses the following [suggestions and feedback](#251) made at ipfs thing. Todo item 3 has been implemented in PR #257. 

We want to limit wnfs dependency on `libipld`. Changing the codec type from `IpldCodec` to `u64` lets us replace `libipld` with `libipld_core` in `wnfs` and `wnfs-wasm` crates. In a way this also removes the limitation of `libipld::IpldCodec` enum. 

We think `bytes::Bytes` is a much more efficient way of getting bytes to and from a BlockStore (for BlockStores implementation that care about that sort of thing). Previously we used `Cow<Vec<u8>>` and we have considered `Read/Write` but `bytes::Bytes` is just more flexible. This change basically makes block storage and retrieval more efficient and flexible while maintaining `no_std` compatibility.

- [x] Replace IPLDCodec with u64.
- [x] Remove `wnfs` dependency on libipld.
- [x] Prefer `bytes::Bytes` to `Vec<u8>` in BlockStore interface. 

## Test plan (required)
-  Testing
    
    ```bash
    scripts/rs-wnfs test
    ```

## Closing issues

Fixes #251

### DIFF
--- a/wnfs-common/Cargo.toml
+++ b/wnfs-common/Cargo.toml
@@ -20,6 +20,7 @@ authors = ["The Fission Authors"]
 anyhow = "1.0"
 async-once-cell = "0.4"
 async-trait = "0.1"
+bytes = { version = "1.4.0", features = ["serde"] }
 chrono = { version = "0.4.23", default-features = false, features = ["clock", "std"] }
 futures = "0.3"
 libipld = { version = "0.16", features = ["dag-cbor", "derive", "serde-codec"] }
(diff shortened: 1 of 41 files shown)


matheus23:
I'd say this should be a `feat` PR, since it changes the signature in one of our exposed traits.

matheus23 (approved):
Nice. Love it.

PR #287, opened 2023-06-21
chore: update READMEs

appcypher:
Update the readme examples and elaborate on some of the filesystem concepts.


matheus23 (approved):
Thanks for noticing these!

PR #305, opened 2023-07-07
feat: Switch from AES-GCM to XChaCha20-Poly1305

matheus23:
## Summary

This PR
- Switches out the `aes-gcm` lib for `chacha20poly1305`
- Switched the `Aes256Gcm` struct for `XChaCha20Poly1305`
- Switched the nonces from `aes::Nonce` to `XNonce`
- Renamed `AesError` into `CryptError` (We still use AES-KW, so a more generic name made sense)
- Removed `aes.rs` file with `AesKey` struct
- Moved some custom key wrap/unwrap code from `PrivateRef::to_serializable` and `PrivateRef::from_serializable` to just use `TemporalKey::key_wrap_{encrypt,decrypt}`

The motivation for switching is simply the extended nonce. It let's us always generate a random nonce for every ciphertext and sleep well at night.
The canonical downside of *not* using AES-based ciphers is performance in case AES native instructions (AESNI) are available. However, this is not our main bottleneck & we'll run in lots of environments (e.g. Wasm) that don't benefit from AESNI.


expede:
Screenshot 2023-07-07 at 11 04 50

wnfs/src/error.rs
```diff
@@ -68,14 +68,14 @@ pub enum ShareError {
     AccessKeyNotFound,
 }
 
-/// AES-GCM errors.
+/// Symmetric encryption errors.
 #[derive(Debug, Error)]
-pub enum AesError {
+pub enum CryptError {
```
expede, line 73: 👍

wnfs/src/private/node/header.rs
```diff
@@ -130,7 +130,7 @@ impl PrivateNodeHeader {
     /// ```
     /// use std::rc::Rc;
     /// use wnfs::private::{
-    ///     PrivateFile, AesKey,
```
expede, line 133: Why did this parameter disappear?

Also, if it exists for Rust, I'd strongly suggest setting up doctests.

expede (approved):
🙌 

Huzzah! LGTM!

wnfs/src/private/node/header.rs
matheus23: We do run doctests. This disappeared because it's an unused import (and I removed the struct in this PR). Doctests is actually how I noticed that.

expede: Oh interesting! Is it just testing that it compiles, or is there an implicit assertion of some kind?

matheus23: It runs the code. This doctest in particular doesn't have an assertion, apart from the fact that if anything in here we're to panic, the doctest would fail, but others do.

expede: Right gotcha — that's the same behaviour in Elixir (it catches any panics and just fails the doctest).

Maybe there are different pattenrs in Rust than Elixir, and it's your code so this is just a suggestion from a non-Rustaciean: the type system will already catch the kind of thing that's written here, so while the documentation side is helpful, it may be _more_ helpful with some broader exploration. I've found that doctests can do a really great job of encouraging people to write great docs when they have to explain why you'd care about the function and/or its output, and often turns into stories about "you care about X, so this gives you back Y which you need for Z".

matheus23: Agreed. The docs could use some work.
I'd actually have preferred if we had *less* doctest while the API is still in flux, and then do a doc pass when it's more stable. It's been a lot of work updating doctests for changes previously.

PR #306, opened 2023-07-10
feat: Switch from SHA3-256 to BLAKE3-256

matheus23:
## Summary

- Switched any `sha3` dependencies out for `blake3` with the `traits-preview` feature. That feature made the switch pretty seamless.
- Switched `prime_hash` impl to using little-endian big integer encoding, otherwise some tests would run out of stack space. `BigUInt::from_bytes_be` actually simply copies over the bytes into a new array that it reverses and runs `BigUint::from_bytes_le` on, which causes quite a lot of allocation.
- Added a test to make sure future changes to `prime_hash` don't unintentionally change hash results.
- Adjusted the way private file blocks are referenced to make it easier to construct the names
- Added `base_name` to `ExternalFileContent`. This allows users with snapshot access only to derive the necessary block labels (they wouldn't be able to access the `header.name`).
- Did some minor changes to `keys.rs`, making the API slimmer (the `SnapshotKey` and `TemporalKey` newtypes are now private), and removing `From<>` impls in favor of documented, (hopefully hard to misuse?) special-purpose functions.

Still having the "tests run out of stack" issue every now and then. After a couple of complicated debugging sessions with `lldb`, it turns out the issue is mostly having functions with lots of await points (e.g. unit tests!). All of the Futures are allocated on the stack, blowing it. There's no good solution that I'm aware of, except for `RUST_MIN_STACK=2500000 cargo test`.


expede:
I haven't actually like plotted the benchmarks or anything, but just looking at the raw numbers it looks like this PR mostly runs as-fast-or-faster than the previous PR 🙌

expede:
Actually, do we have a tool that plots the benches set up? Save me from tabbing back and forth between the plaintext results 😛

matheus23:
We have this: https://wnfs-wg.github.io/rs-wnfs/dev/bench/

But I don't think we have anything like that for branches.

expede:
Oh these benchmarks are for native code though, right? Do we have Wasm benches?

matheus23:
No WASM benches currently :/

expede:
(I really dislike the lack of threading in main thread PR comments)

We should at minimum run some manual tests for this PR I think. Setting up Wasm benches would be nice in general (we have wasm-js tests)

expede:
Was chatting with @zeeshanlakhani about this, and he says that the right way to bench Wasm with using JS wrappers

appcypher (approved):
<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

PR #310, opened 2023-07-19
Fix `search_latest` on the root directory and implement `searchLatest` in wasm bindings

### LINKED ISSUE #309: Wasm: Expose `.searchLatest` on `PrivateNode`
Perhaps it's worth considering having a `search_latest: bool` param on `PrivateNode::load`, since otherwise people might miss that that's required.

matheus23:
The title pretty much covers it.

This PR also includes a previously failing test case, and a wasm test case testing the new API.

Fixes #309


icidasset (approved):

appcypher (approved):
Looks good to me

PR #314, opened 2023-07-24
chore: data format snapshot tests

### LINKED ISSUE #249: Unique canonical "entrypoint" API for private WNFS

appcypher:
This PR adds snapshot tests for the data format of WNFS.
The tests are identifiable by their encapsulating module, `snapshot_tests`.

- [x] Snapshot tests for wnfs data format

## Test plan (required)
-  Snapshot tests  
    ```
    cargo install cargo-insta
    ```
    
    ```bash
    cargo insta test --check -- snapshot_tests    
    ```

-  Testing
    
    ```bash
    scripts/rs-wnfs test
    ```


wnfs-common/src/utils/test.rs
```diff
@@ -37,51 +46,64 @@ pub trait Sampleable {
...
-        // Then we deserialize it back to `Ipld`. We can be sure value based on the generated cbor.
-        let value: Ipld = decode(&cbor, DagCborCodec)?;
-
-        let json = encode(
-            &ipld_serde::to_ipld(MockData { value, cbor })?,
-            DagJsonCodec,
-        )?;
-
-        self.put_block(json, CODEC_DAG_JSON).await
+    pub fn get_values(&self) -> Result<Vec<BlockSnapshot>> {
```
matheus23, line 64: _Can_ we make this return `BTreeMap<String, BlockSnapshot>` instead of `Vec<BlockSnapshot>`? 
The `String`s would be the encoded CIDs. This way we can follow the links in the snapshot.
(Ideally the snapshot is also topologically sorted, but... that's perhaps asking for a lot for the initial version :sweat_smile: )

appcypher:
Deciding to push exhaustive tree history snapshotting the into another PR.

matheus23:
Wasm tests error out weirdly:
```
error[E0433]: failed to resolve: use of undeclared crate or module `imp`
Error:   --> /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wait-timeout-0.2.0/src/lib.rs:66:9
   |
66 |         imp::wait_timeout(self, dur)
   |         ^^^ use of undeclared crate or module `imp`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `wait-timeout` (lib) due to previous error
Error: warning: build failed, waiting for other jobs to finish...
❌ [compile] Failed with exit status 101
```

appcypher:
The suggestions given by the nightly linter are incorrect and I'm not sure what is causing it.

matheus23:
Hmm. Related to this? https://github.com/rust-lang/rust-clippy/issues/11179

Nightly should be the *latest latest* clippy though, right? That issue seems to be fixed already. Is it a clippy regression?

matheus23:
If we can't manage to fix the issue, can we add a bunch of `allow` annotations so we get rid of these errors?

appcypher:
Or we could ignore the errors since it is likely going to fix itself in the future.
We don't want to sprinkle allows in the code for sth that isn't a real problem.

wnfs/src/private/node/serializable.rs
```diff
@@ -52,12 +52,11 @@ pub(crate) struct PrivateNodeHeaderSerializable {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub(crate) struct PrivateRefSerializable {
-    #[serde(serialize_with = "crate::utils::serialize_byte_slice32")]
-    #[serde(deserialize_with = "crate::utils::deserialize_byte_slice32")]
+    #[serde(with = "serde_byte_array")]
     pub label: HashOutput,
     #[serde(rename = "snapshotKey")]
     pub snapshot_key: SnapshotKey,
-    #[serde(rename = "temporalKey")]
```
matheus23, line 59: Accidental rename removal :sweat_smile:

matheus23: May be worth just switching to that one struct-level directive that renames everything into camel-case. But we can do that in another PR, this way we could make sure that it doesn't actually change snapshots ;)

matheus23:
> Or we could ignore the errors since it is likely going to fix itself in the future.
> We don't want to sprinkle allows in the code for sth that isn't a real problem.

Hmmm. As long as these directives are very local, I think they're not too bad.
I don't want us to get used to the :x: on `main`. We'd have to check exactly which error it is in the logs everytime to make sure we're not ignoring some error that's more important.

If you think about it: In both cases we're ignoring errors. In one case we're ignoring them for the whole project, in the other we're ignoring them for only a couple of methods.
Perhaps we can set us a reminder to check back in a month to see if we can remove the `allow` annotations?

appcypher:
`needless_pass_by_mut_ref` does not exist on stable yet.

matheus23:
Ugh. That's so annoying

wnfs-common/src/utils/test.rs
```diff
@@ -0,0 +1,127 @@
...
+//--------------------------------------------------------------------------------------------------
+// Type Definitions
+//--------------------------------------------------------------------------------------------------
+
+type CidHandler = Box<dyn Fn(&Bytes) -> Result<Bytes>>;
+
+#[derive(Default)]
+pub struct MockStore {
+    inner: MemoryBlockStore,
+    cid_handlers: HashMap<Cid, CidHandler>,
```
matheus23, line 31: Hm. Maybe also rename this. I may have originally suggested `CidHandler` (I don't remember!), but it's not actually handling CIDs, it's handling the blocks. Perhaps we call it BlockHandler?

wnfs-hamt/src/node.rs
```diff
@@ -753,7 +754,7 @@ impl<K, V, H: Hasher> Node<K, V, H> {
         K: Serialize,
         V: Serialize,
     {
-        let bitmask_ipld = ipld_serde::to_ipld(self.bitmask.as_raw_slice())?;
+        let bitmask_ipld = ipld_serde::to_ipld(ByteArray::from(self.bitmask.into_inner()))?;
```
matheus23, line 757: :sparkles: :heart:

wnfs-hamt/src/hamt.rs
```diff
@@ -223,17 +223,41 @@ where
...
+
+#[cfg(test)]
+mod snapshot_tests {
+    use super::*;
+    use wnfs_common::utils::MockStore;
+
+    #[async_std::test]
+    async fn test_hamt() {
+        let store = &MockStore::default();
+        let node = &mut Rc::new(Node::<[u8; 4], String>::default());
```
matheus23, line 250: Nit: Can we simply do `Node::<u32, String>` here?
The `[u8; 4]` makes the snapshot read ... just slightly more weird and confusing.

wnfs-nameaccumulator/Cargo.toml
```diff
@@ -31,8 +31,12 @@ thiserror = "1.0"
...
 [dev-dependencies]
+async-std = { version = "1.12.0", features = ["attributes"] }
 hex = "0.4.3"
+insta = { version = "1.31.0", features = ["json"] }
 proptest = "1.1"
 rand = "0.8.5"
 rand_chacha = "0.3"
+serde_json = "1.0.103"
 test-strategy = "0.3"
+wnfs-common = { path = "../wnfs-common", features = ["test_utils"] }
```
matheus23, line 42: Yep, this is the way. No `version` field in dev-dependencies.

wnfs-nameaccumulator/src/name.rs
```diff
@@ -629,18 +629,15 @@ mod tests {
...
         let segment = NameSegment::new(rng);
 
-        let ipld = libipld::serde::to_ipld(&segment).unwrap();
-        let mut bytes = Vec::new();
-        ipld.encode(DagCborCodec, &mut bytes).unwrap();
-
-        let ipld = Ipld::decode(DagCborCodec, &mut Cursor::new(bytes)).unwrap();
-        let segment_back = libipld::serde::from_ipld::<NameSegment>(ipld).unwrap();
+        let bytes = encode(&segment, DagCborCodec).unwrap();
+        let segment_back: NameSegment = decode(&bytes, DagCborCodec).unwrap();
```
matheus23, line 640: :ok_hand: :pray:

wnfs-nameaccumulator/wnfs-nameaccumulator-2.code-workspace
```diff
@@ -0,0 +1,8 @@
+{
+  "folders": [
+    {
+      "path": "."
+    }
+  ],
+  "settings": {}
+}
```
matheus23, line 8: Is this an accidentally committed file?

wnfs/Cargo.toml
```diff
@@ -28,31 +28,40 @@ bytes = "1.4.0"
...
+serde_ipld_dagcbor = "0.4"
+sha3 = "0.10"
+skip_ratchet = { version = "0.3", features = ["serde"] }
 thiserror = "1.0"
-wnfs-common = { path = "../wnfs-common", version = "0.1.23" }
-wnfs-hamt = { path = "../wnfs-hamt", version = "0.1.23" }
-wnfs-nameaccumulator = { path = "../wnfs-nameaccumulator", version = "0.1.23" }
+wnfs-common = { path = "../wnfs-common", version = "0.1" }
+wnfs-hamt = { path = "../wnfs-hamt", version = "0.1" }
+wnfs-nameaccumulator = { path = "../wnfs-nameaccumulator", version = "0.1" }
```
matheus23, line 47: Hm... I'm not sure about this.
What's the reasoning for going for less precise version bounds here? We're not really following semver, so those bounds aren't actually accurate. In fact, we should probably change the version bounds to `version = "= 0.1.23"` because it literally fails to compile otherwise.

wnfs/src/private/file.rs
```diff
@@ -85,20 +85,16 @@ pub(crate) struct PrivateFileContent {
 /// The content of a file.
 /// It is stored inline or stored in blocks.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
matheus23, line 88: :sparkles:

wnfs/src/private/node/keys.rs
```diff
@@ -54,19 +54,11 @@ pub(crate) const SNAPSHOT_KEY_DSI: &str = "wnfs/snapshot key deriv from temporal
 
 /// The key used to encrypt the content of a node.
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
-pub struct SnapshotKey(
-    #[serde(serialize_with = "crate::utils::serialize_byte_slice32")]
-    #[serde(deserialize_with = "crate::utils::deserialize_byte_slice32")]
-    pub(crate) [u8; KEY_BYTE_SIZE],
-);
+pub struct SnapshotKey(#[serde(with = "serde_byte_array")] pub(crate) [u8; KEY_BYTE_SIZE]);
```
matheus23, line 57: :pray: :sparkles:

wnfs/src/private/directory.rs
```diff
@@ -2170,3 +2170,57 @@ mod tests {
...
+        let base_name = forest.empty_name();
+        let paths = [
+            vec!["text.txt".into()],
+            vec!["music".into(), "jazz".into()],
+            vec!["videos".into(), "movies".into(), "anime".into()],
+        ];
+
+        let root_dir = &mut Rc::new(PrivateDirectory::new(
+            &base_name,
+            DateTime().fake_with_rng(rng),
```
matheus23, line 2197: Very minor nit: This results in a bunch of negative times. I don't know if we want to specifically support negative timestamps forever and we also don't really need to generate random ones here. What do you think of just a default time everywhere? Jan 1st 1970 or some other fun date.

wnfs/src/snapshots/wnfs__root_tree__snapshot_tests__root_filesystems.snap
```diff
@@ -0,0 +1,571 @@
...
+            "bytes": "6iFLiInnD8bvPCKqoCh6VJBjscDFlolx6vW76ycbN40"
+          }
+        },
+        "smallCounter": 239
+      }
+    },
+    "cbor": "o2dpbnVtYmVyWCDbJjulZAQRPwvLwGUS/Bs+0LEYcNFrz5lrto+e1bvt82dyYXRjaGV0pmRzYWx0WCC7KHzq6kw1G+8na9pNAFdGT7bXG+qRDt5UyozM4CZZcmVsYXJnZVggoaoqYCumskEku0VUn5FpFnCqV1M4VER0qgrTHHRxbTBmbWVkaXVtWCC+LXPTLYA2gfUzownXDZXtd0bXNN7cbzIWnJ8XHS+rMWVzbWFsbFgg6iFLiInnD8bvPCKqoCh6VJBjscDFlolx6vW76ycbN41tbWVkaXVtQ291bnRlchhObHNtYWxsQ291bnRlchjvZG5hbWVZAQBfm2qUNg6ZpSkL3ahC2oyZAHEHMTsYdbs4dtipMMy7+hnfHTs+6lHrMiwLxGT9G7i/wArD9CKp7ABCMYmC0weSZaPWJrXIIwynKwyfLEHIOVdj6PlvSKNs6GvVDqd6MPQSwBx/8KhpWtEFjK1ALl3VIuJg+0YWJhW1I69JJAJVdxjonoMaO3V9dqMom0OrEfDwLbOSKia2tls2YZOKgio77IXRNyjl+A3YV4ertUmNY0tgVFtmntZrgawU8toe5lGhA4nfRG9lsS6Xn0UbvTugqOuOABEo8pSp/gxYizuuTaGvDXr0awNBSig9Os8N3N6tdLwNdK12k4+hc1JJUiCn"
+  },
+  "bafkr4idh7nmqq4qaiphdru4eimfwq3fmj3vktxi4fzqf2yjtjrv5o6ignm": {
+    "value": "ello wor",
```
matheus23, line 217: Hm. This seems weird to me :thinking: 
Shouldn't this be "hello world"?

matheus23 (approved):
Awesome stuff! Thanks so much.

There's some nits, some naming suggestions, one weird output, but I trust you with the rest :)

wnfs/Cargo.toml
appcypher: It fails to compile? I remember zeeshan suggesting keeping patch flexible for all things libraries (can't find the quote right now). That is why our deps have just major.minor versions.

matheus23: Yep I know. But we make breaking changes between e.g. 0.1.21 and 0.1.23 versions.
We recently had [someone in the discord](https://discord.com/channels/478735028319158273/877623827331428403/1136628690147618899) who hit an issue with wnfs 0.1.21 depending on wnfs-common 0.1.23.

matheus23: Eventually we should just get out of version 0.1 & release something that's 1.0 and start doing semver checks.

wnfs-hamt/src/hamt.rs
matheus23: Ah. Right. I regret having the `AsRef<[u8]>` requirement on node's `K` :/
It made it impossible to simply use `NameAccumulator` for `K`, without adding a `serialized_cache: OnceCell<[u8; 256]>` to it.
We should probably add a small trait `HamtKey` or `Hashable` or something like that. We can then additionally have a `impl<T: AsRef<[u8]>> Hashable for T` if we want to, but it'll be possible to implement `Hashable` for `u32` and `NameAccumulator` (without a `serialized_cache`).

wnfs-common/src/utils/test.rs
```diff
@@ -0,0 +1,131 @@
...
+    block_handlers: HashMap<Cid, BlockHandler>,
+}
+
+base64_serde_type!(Base64Standard, base64::engine::general_purpose::STANDARD);
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct BlockSnapshot {
+    pub value: Value,
+    #[serde(with = "Base64Standard")]
+    pub cbor: Vec<u8>,
```
matheus23, line 40: Sorry, just one more minor nit, can we call this "bytes", now that sometimes it's not CBOR, but e.g. encrypted bytes:
```suggestion
    pub bytes: Vec<u8>,
```

PR #317, opened 2023-07-25
fix: Improve performance of `get_revision_name()`

matheus23:
This simply caches the result of `get_revision_name()`, as that's accessed quite a lot and would otherwise cause lots of repeated BigUint operations, which are costly.

It improves performance of general filesystem operations by roughly 20% in my limited testing (I compared runtime of `cargo test --lib -p wnfs -- "private::directory::"`).


appcypher (approved):
LGMT! 🎉

Issue #61, opened 2022-09-22
Simplify wasm-pack logic


matheus23:
We don't necessarily need to have `--weak-refs` flag support in `wasm-pack`, we can make wasm-bindgen generate weak-ref stuff via environment variables: https://github.com/rustwasm/wasm-pack/pull/937#issuecomment-1254786874

PR #318, opened 2023-07-27
chore: Update JS binding and packaging

bgins:
## Summary

This PR implements the following features:

- [x] Switch to `wasm-bindgen` to generate bindings for node, web, and bundler targets
- [x] Add `wireit` to compile, optimize, and bind for multiple targets
- [x] Update Webpack tests to use the `bundler` output
- [x] Update CI scripts to install `wasm-bindgen` (instead of `wasm-pack`) and `binaryen` (for `wasm-opt`)

This PR updates JS packaging and binding to make the package easier to load in bundlers, node, and Cloudflare Workers. 

It's not perfect! We are testing out improvements over to support SvelteKit and other environments over in https://github.com/fission-codes/wasm-js. But it should be a good starting point, and it will help us with testing in the ODD SDK.

Note that the `web` target is a special case for Cloudflare workers. Excepting node, most other environments will use the `bundler` target.

## Test plan (required)

Run the Webpack test suite. The tests should still work and pass. 👌 

## Closing issues

None yet. Happy to open one.

## After Merge

- [ ] Does this change invalidate any docs or tutorials? The READMEs have been updated. Anything else to do?
- [ ] Does this change require a release to be made? Yes. But will leave that to maintainers. 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,10 +81,10 @@ This library is designed with WebAssembly in mind. You can follow instructions o
 
   - [rust-analyzer][rust-analyzer-guide] is the go-to IDE tool for Rust and if you have it set up, you may want to set the `rust-analyzer.cargo.target` [setting][vscode-settings] to `wasm32-unknown-unknown`
 
-  - Install wasm-pack
+  - Install wasm-bindgen
 
     ```bash
-    cargo install wasm-pack
+    cargo install wasm-bindgen
     ```
 
   - Install [playwright][playwright-guide] binaries
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ This library is designed with WebAssembly in mind. You can follow instructions o
   - Install wasm-bindgen
 
     ```bash
-    cargo install wasm-bindgen
+    cargo install wasm-bindgen-cli
     ```
 
   - Install [playwright][playwright-guide] binaries
--- a/scripts/rs-wnfs.sh
+++ b/scripts/rs-wnfs.sh
@@ -135,7 +135,7 @@ build_accumulator() {
 build_wasm() {
     display_header "💿 | BUILDING WNFS-WASM PROJECT | 💿"
     cd $script_dir/../wnfs-wasm
-    WASM_BINDGEN_WEAKREF=1 npm run build
+    npm run build
 	sed -i.bak \
         -e 's/"name": "wnfs-wasm"/"name": "wnfs"/g' \
         package.json
--- a/scripts/rs-wnfs.sh
+++ b/scripts/rs-wnfs.sh
@@ -135,12 +135,11 @@ build_accumulator() {
 build_wasm() {
     display_header "💿 | BUILDING WNFS-WASM PROJECT | 💿"
     cd $script_dir/../wnfs-wasm
-    WASM_BINDGEN_WEAKREF=1 wasm-pack build --target web
+    WASM_BINDGEN_WEAKREF=1 npm run build
 	sed -i.bak \
-        -e 's/"name": "wnfs-wasm"/"name": "wnfs",\n  "type": "module"/g' \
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -81,10 +81,10 @@ This library is designed with WebAssembly in mind. You can follow instructions o
 
   - [rust-analyzer][rust-analyzer-guide] is the go-to IDE tool for Rust and if you have it set up, you may want to set the `rust-analyzer.cargo.target` [setting][vscode-settings] to `wasm32-unknown-unknown`
 
-  - Install wasm-pack
+  - Install wasm-bindgen
 
     ```bash
-    cargo install wasm-pack
+    cargo install wasm-bindgen
```
matheus23, line 87: Should this be `wasm-bindgen-cli` perhaps?
```suggestion
    cargo install wasm-bindgen-cli
```

wnfs-wasm/README.md
```diff
@@ -17,10 +17,10 @@ WNFS file trees can serialize and be deserialized from [IPLD graphs][ipld-spec]
 
 ## Setting up the Project
 
-- Install `wasm-pack`
+- Install `wasm-bindgen`
 
   ```bash
-  cargo install wasm-pack
+  cargo install wasm-bindgen
```
matheus23, line 23: Here too?
```suggestion
  cargo install wasm-bindgen-cli
```

scripts/rs-wnfs.sh
```diff
@@ -135,12 +135,11 @@ build_accumulator() {
 build_wasm() {
     display_header "💿 | BUILDING WNFS-WASM PROJECT | 💿"
     cd $script_dir/../wnfs-wasm
-    WASM_BINDGEN_WEAKREF=1 wasm-pack build --target web
+    WASM_BINDGEN_WEAKREF=1 npm run build
```
matheus23, line 138: We're now running `wasm-bindgen` ourselves, so we can get rid of this env variable and instead pass `wasm-bindgen --weak-refs` directly.
```suggestion
    npm run build
```

wnfs-wasm/package.json
```diff
@@ -1,10 +1,133 @@
...
+        "TARGET_DIR": {
+          "external": true
+        }
+      },
+      "dependencies": [
+        "compile"
+      ]
+    },
+    "bindgen:bundler": {
+      "command": "wasm-bindgen --weak-refs --target bundler --out-dir dist/bundler ../target/wasm32-unknown-unknown/$TARGET_DIR/wnfs_wasm.wasm",
```
matheus23, line 75: > `wasm-bindgen --weak-refs`
Nice :ok_hand:

matheus23 (approved):
After installing `wasm-opt` and `wasm-bindgen` this is working for me :)
It's slightly unfortunate that installing `wasm-bindgen` is somewhat hard on NixOS today (`cargo install wasm-bindgen-cli` doesn't work because it needs openssl library access & the native nixpkgs package is wasm-bindgen version 0.2.84, but we need 0.2.87 and it complains if they're not matching), but luckily wasm-bindgen is a really portable executable, so I could simply copy it over from where I had it before.

Also, I think this
Fixes #61

walkah:
> It's slightly unfortunate that installing `wasm-bindgen` is somewhat hard on NixOS today (`cargo install wasm-bindgen-cli` doesn't work because it needs openssl library access & the native nixpkgs package is wasm-bindgen version 0.2.84, but we need 0.2.87 and it complains if they're not matching), but luckily wasm-bindgen is a really portable executable, so I could simply copy it over from where I had it before.

sounds like we need some `devShells` in here :)

README.md
bgins: Yep! Good catch. 🙌

PR #319, opened 2023-08-01
feat: Implement public directory cp & more efficient copy for `PrivateFile`

### LINKED ISSUE #316: Wasm: Add bindings for `cp` in `PrivateDirectory`
Another missing binding

matheus23:
Fixes #316 (although what was missing was actually `cp` for *Public* directory, not private)

Since #306 we can actually do a more efficient copy of external `PrivateFile` content that doesn't require re-encryption, because we can directly refer to the old content (using the `bare_name` property of `ExternalFileContent`).

I've decided to set this as the default copy algorithm. There's no benefit to re-encrypting the actual content bytes. You already get the key rotation on the wrapping `PrivateFile`, and it doesn't implicitly give you access to anything else besides the bytes you're supposed to be able to read.

I think this supersedes #256 at least in terms of what Banyan needs.

### DIFF
--- a/wnfs-wasm/src/fs/public/directory.rs
+++ b/wnfs-wasm/src/fs/public/directory.rs
@@ -211,6 +211,30 @@ impl PublicDirectory {
         }))
     }
 
+    /// Copies a specific node to another location.
+    pub fn cp(
+        &self,
+        path_segments_from: &Array,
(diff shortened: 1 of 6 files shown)


appcypher (approved):
🚀🚀🚀

PR #326, opened 2023-08-11
fix: More reliably cache `NameAccumulator` modexps

matheus23:
`BigUint::modexp` operations take up to 14ms in Wasm on my machine, and it's the core operation for `NameAccumulator`s.

This introduces a LRU cache inside `HamtForest` that caches `NameAccumulator` operations.

This is much more reliable than sprinkling `OnceCell`s every here and there, since
1. The "eviction" of these `OnceCell` caches depends on the lifetime of the object they're contained in, and it's harder to make these structs longer-lived in some cases.
2. Sometimes the structs that these `OnceCell`s are contained inside are mutable via `&mut self` operations. We need to remember to always empty these `OnceCell`s when that happens. This is bad for two reasons: (1) we might forget doing so sometimes, keeping stale, incorrectly cached data around and (2) even though the cache doesn't match the object anymore, it may still be a useful cache!

### DIFF
--- a/wnfs/src/private/node/header.rs
+++ b/wnfs/src/private/node/header.rs
@@ -226,6 +226,8 @@ impl PrivateNodeHeader {
 
 impl PartialEq for PrivateNodeHeader {
     fn eq(&self, other: &Self) -> bool {
-        self.inumber == other.inumber && self.ratchet == other.ratchet && self.name == other.name
+        // We skip equality-checking the name, since it depends on where the node header was mounted.
+        // The inumber should suffice as identifier, the ratchet covers the revision part.
+        self.inumber == other.inumber && self.ratchet == other.ratchet
     }
 }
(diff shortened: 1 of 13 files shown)


wnfs/src/private/node/header.rs
```diff
@@ -173,21 +167,23 @@ impl PrivateNodeHeader {
 
     /// Encrypts this private node header in an block, then stores that in the given
     /// BlockStore and returns its CID.
-    pub async fn store(&self, store: &impl BlockStore, setup: &AccumulatorSetup) -> Result<Cid> {
+    ///
+    /// This *does not* store the block itself in the forest, only the the given block store.
```
appcypher, line 171: There is a typo here.

appcypher (approved):
This looks great to go. Another performance boost. 
Great stuff @matheus23. We slowly getting there.

PR #329, opened 2023-08-15
chore: Release 0.1.24

matheus23:
Also ignoring the `needless_pass_by_ref_mut` lint.
That lint has a bug that causes false positives. It'll be fixed when this is released: https://github.com/rust-lang/rust-clippy/pull/11314
However, it's a nightly lint, so unfortunately I can't ignore it via a `#[allow(clippy::needless_pass_by_ref_mut)]`, since that makes *stable* clippy fail with an unknown lint error.
So I made it ignore all lints from the `suspicious` group on the affected functions.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,5 @@
 {
   "wnfs": "0.1.23",
-  "wnfs-bench": "0.1.23",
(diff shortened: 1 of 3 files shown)


zeeshanlakhani (approved):
hope the release works!

PR #340, opened 2023-08-28
feat: Implement storing externally encrypted content in `Metadata`

matheus23:
This work enables File Variants (https://github.com/wnfs-wg/roadmap/issues/7).

Essentially it exposes the struct used for storing pointers & keys to content encrypted inside the `PrivateForest` as well as some utilities to store these in a file's (or directory's) `Metadata`.

See also the new example `file_variants.rs`.

### DIFF
--- a/wnfs/examples/file_variants.rs
+++ b/wnfs/examples/file_variants.rs
@@ -46,7 +46,7 @@ async fn main() -> Result<()> {
 
     // We store the content in the file metadata.
     // This will update the `file: Rc<PrivateFile>` for us with a new reference.
-    content.store_in_metadata(file.get_metadata_mut()?, "thumbnail")?;
+    content.store_in_metadata(file.get_metadata_mut_rc()?, "thumbnail")?;
 
     // We store the new reference in the forest.
     file.as_node().store(forest, store, rng).await?;
--- a/wnfs/examples/file_variants.rs
+++ b/wnfs/examples/file_variants.rs
@@ -48,8 +48,6 @@ async fn main() -> Result<()> {
     // This will update the `file: Rc<PrivateFile>` for us with a new reference.
     content.store_in_metadata(file.get_metadata_mut()?, "thumbnail")?;
 
-    println!("{:#?}", file.get_metadata().get("thumbnail"));
-
     // We store the new reference in the forest.
     file.as_node().store(forest, store, rng).await?;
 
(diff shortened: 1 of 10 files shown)


wnfs/src/private/file.rs
```diff
@@ -813,6 +655,264 @@ impl PrivateFileContent {
...
+        Ok(match wrapped {
+            MetadataContentCapsule::PrivateForestContent(content) => content,
+        })
+    }
+
+    /// Store these pointers & key to some private forest content in given metadata under given key.
+    pub fn store_in_metadata(&self, metadata: &mut Metadata, key: &str) -> Result<()> {
+        metadata.put_serializable(key, MetadataContentCapsule::PrivateForestContent(&self))?;
+        Ok(())
+    }
```
appcypher, line 772: I'm thinking this function really belongs to PrivateFile. So we wouldn't need separate fns like `get_metadata_mut_rc` that if missed can lead to catastrophic failures.
A function like `PrivateFile::upsert_metadata`, perhaps an Entry API. The point is to let file handle updating the metadata since it is the custodian and it best knows what it needs to be in the correct state.

wnfs/examples/file_variants.rs
```diff
@@ -0,0 +1,63 @@
...
+        b"secondary content".to_vec(),
+        forest,
+        store,
+        rng,
+    )
+    .await?;
+
+    // We store the content in the file metadata.
+    // This will update the `file: Rc<PrivateFile>` for us with a new reference.
+    content.store_in_metadata(file.get_metadata_mut_rc()?, "thumbnail")?;
```
appcypher, line 49: Following the suggestion above, this would then read more naturally as:

```
file.upsert_metadata("thumbnail", |_| content.as_metadata_value());
```

What do you think?

appcypher (approved):
File variants ftw! Great work. I like the example as well.
I only have a suggestion concerning the `store_in_metadata` method.

<div align="center">
    <a href="#" target="_blank">
        </img>
    </a>
</div>

wnfs/examples/file_variants.rs
matheus23: I like the idea of an `as_metadata_value` function :)

I don't want to re-introduce functions like `upsert_metadata` that really should belong to the `Metadata` implementation, and partially exists there already (`put`, `get`, `delete`, etc.).

It's somewhat annoying at the moment that sometimes you have a `&mut PrivateFile` and sometimes you have a `&mut Rc<PrivateFile>` (and we don't expose `prepare_next_revision`), but :man_shrugging: not sure about a better solution there in general.

appcypher: You are right. I can't come up with a fine solution for this.

PR #341, opened 2023-08-30
fix: Fix `search_latest` on `write` and file mountpoints

matheus23:
This fixes two bugs reported by @icidasset:
- We don't respect the `search_latest` parameter during calls like `write` or `mkdir` for the root directory.
- The mountpoint was incorrectly set in `search_latest_node` to the empty name accumulator in some cases.

### DIFF
--- a/wnfs/src/private/directory.rs
+++ b/wnfs/src/private/directory.rs
@@ -304,6 +304,10 @@ impl PrivateDirectory {
         forest: &impl PrivateForest,
         store: &impl BlockStore,
     ) -> Result<SearchResult<&'a mut Self>> {
+        if search_latest {
+            *self = self.clone().search_latest(forest, store).await?;
+        }
+
         let mut working_dir = self.prepare_next_revision()?;
         for (depth, segment) in path_segments.iter().enumerate() {
             match working_dir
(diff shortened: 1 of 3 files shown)


wnfs/src/private/directory.rs
```diff
@@ -304,6 +304,10 @@ impl PrivateDirectory {
         forest: &impl PrivateForest,
         store: &impl BlockStore,
     ) -> Result<SearchResult<&'a mut Self>> {
+        if search_latest {
+            *self = self.clone().search_latest(forest, store).await?;
+        }
+
```
matheus23, line 310: This fixes the first issue

wnfs/src/private/node/node.rs
```diff
@@ -413,7 +413,7 @@ impl PrivateNode {
         store: &impl BlockStore,
     ) -> Result<Vec<PrivateNode>> {
         let header = self.get_header();
-        let mountpoint = header.name.parent().unwrap_or_else(|| forest.empty_name());
+        let mountpoint = header.name.parent();
```
matheus23, line 416: This fixes the second issue.

icidasset (approved):
Remote high five!

Issue #315, opened 2023-07-24
Share API improvments


matheus23:
@appcypher and I had some ideas on how we could improve the share API.

Consider this a list of "may or *may not* be a good idea".

- We should probably provide a (feature-gated) implementation of `ExchangeKey`. Feature-gated, because pulling in the RSA dependency may be undesired for WASM, where we might instead want to simply depend on WebCrypto (but in that case this needs re-thinking anyways, since that abstraction would require passing in an object from the JS side, not a static trait).
- `RsaError` should probably be renamed to something more abstract when we add other protocols (e.g. Ed25519)
- We should compute `find_current_share_counter` for the user (by default, at least). It can be made faster by caching previous results, but that doesn't mean the default case needs to be a more complicated API.
- We probably don't need the `Share` struct. The `share` function is just fine.
- Perhaps it's also possible to remove the `receiver` and `sender` modules, and simply merge them all into the `share` module.

icidasset:
Currently it's necessary to keep a global counter because of some issues with the API. The `share()` function expects a `counter` but a counter depends on a public key and a DID, the function however retrieves the public keys for you; so it would make sense if the `share` function also figured out the counter for you. The counter should be returned to the use.

That said, I think the counter should be the same for all the public keys in the same share. Reason being that the user will need to use the counter to retrieve older shares (ie. not the latest) (retrieving a share automatically uses the last counter).

Also, `fetch_exchange_keys` is missing on the wasm side.

PR #343, opened 2023-08-30
feat: Remove `Share` struct, add documentation, add `rc` constructor variants

appcypher:
## Summary

This PR adds and updates existing documentation, doc tests and makes other API clean ups.

- [x] Add more information to docs home page
- [x] Remove unused types and Share struct
- [x] Update READMEs
- [x] Add rc constructor variants

## Test plan (required)
-  Testing the code.
    
    ```bash
    scripts/wnfs.sh test
    ```

### DIFF
--- a/wnfs/src/lib.rs
+++ b/wnfs/src/lib.rs
@@ -19,7 +19,7 @@
 //! #[async_std::main]
 //! async fn main() -> Result<()> {
 //!     // Create a new public directory.
-//!     let dir = &mut PublicDirectory::rc(Utc::now());
+//!     let dir = &mut PublicDirectory::new_rc(Utc::now());
 //!
 //!     // Create an in-memory block store.
 //!     let store = &MemoryBlockStore::default();
@@ -67,10 +67,10 @@
 //!     let rng = &mut thread_rng();
 //!
 //!     // Create a private forest.
-//!     let forest = &mut HamtForest::rc_trusted(rng);
+//!     let forest = &mut HamtForest::new_trusted_rc(rng);
 //!
 //!     // Create a new private directory.
-//!     let dir = &mut PrivateDirectory::rc(&forest.empty_name(), Utc::now(), rng);
+//!     let dir = &mut PrivateDirectory::new_rc(&forest.empty_name(), Utc::now(), rng);
 //!
 //!     // Add a file to /pictures/cats directory.
 //!     dir.mkdir(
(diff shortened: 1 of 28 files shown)


wnfs/src/lib.rs
```diff
@@ -4,8 +4,122 @@
...
+//! }
+//! ```
+//!
+//! This example introduces a few new concepts. The first is the `HamtForest` which is a HAMT that can contain multiple file trees and implements the `PrivateForest` interface needed for persisting private file systems.
+//!
+//! The second is the `Name` (returned from `forest.empty_name()`) and `NameAccumulator` that lets us identify nodes in the filesystem, and are suitable for offspring proving.
+//!
+//! Finally, we have the random number generator, `rng`, that the library uses for generating new keys and other random values needed for the protocol.
+//!
+//! Check the [`wnfs/examples/`](wnfs-examples) folder for more examples.
```
matheus23, line 113: Barney thumbs up Gif

matheus23 (commented):
Big fan of most changes, just not a big fan of saving a bunch of characters with the `rc` name.
Wdyt about `new_rc`? Or perhaps an `rc` function with the signature `fn rc(self) -> Rc<Self>)`? So it's `&mut PrivateDirectory::new(...).rc()` (perhaps that's too easy to miss in examples though)?

appcypher:
> Or perhaps an rc function with the signature fn rc(self) -> Rc<Self>? So it's &mut PrivateDirectory::new(...).rc() (perhaps that's too easy to miss in examples though)?

Thought about making an `rc` function; probably should go with that.
I know this change adds little but I think it simplifies the interface makes it less verbose. We almost always wrap these types in Rcs. They have little use outside of that so I think it only makes sense to make it part of the API.
Alsop it is in line with functions like `Box::pin` that wraps a Box in Pin. There are similar functions like this in the ecosystem.

matheus23:
> Also, it is in line with functions like `Box::pin` that wraps a Box in Pin. There are similar functions like this in the ecosystem.

hmm I see. I'd argue `Box` is just inherently a simpler type (& `Box::new` thus a simpler function), since it's just a wrapper.
The constructor for `PrivateDirectory::new` takes multiple arguments and is more complex than a wrapper.

What do you think of calling the function `new_rc`?

Also. Perhaps worth trying this out: How crazy would it be if `new` simply always returned an `Rc`? Do we have code in rs-wnfs that would break? Maybe we can base our whole API around `Rc<PrivateDirectory>` and `&mut Rc<PrivateDirectory>`...

matheus23 (approved):
Addresses some of the #315

appcypher:
@matheus23 Thanks for fixing 👍. Was going to get to it today.

PR #347, opened 2023-09-04
chore: Remove release please & bump versions

matheus23:
I want to move away from release-please because:
- It continues trying to release `wnfs-bench`, even though it shouldn't. (We've tried removing `wnfs-bench` from all configuration JSONs, but release-please just adds it back, presumably because it reads it from the history somewhere)
- It doesn't handle new crates in the workspace well at all. (Between removing `wnfs-namefilter` and adding `wnfs-nameaccumulator` it just got totally confused and we had to manually publish the *second* release of `wnfs-nameaccumulator`. This also caused all other crates to not release properly, since they depend on `wnfs-nameaccumulator`...)
- Now that we've done one manual release, it's gotten completely confused. (See #344 and the CHANGELOGs it generated)
- While putting together this PR, I realized that the CHANGELOG formatting got messed up at some point for the `wnfs-hamt` crate.

So this PR simply removes the workflow.
I also bumped the versions in preparation for a release & added CHANGELOGs.


matheus23:
Yep. Back to manual :/

PR #348, opened 2023-09-10
fix: Don't drop data sometimes during re-serialization in wnfs-hamt

matheus23:
This fixes a bug reported by @ehsan6sha where the wnfs-hamt implementation would sometimes drop data, when a data structure would be used in-memory after serialization, then deserialized, operated on and finally used again.

I also modified all proptests to check for bugs with serialization roundtripping and added another proptest, as well as a proptest regression file.


matheus23:
Yeah, it's slower, but also more correct.
And we're talking on the order of microseconds, so not a bottleneck usually.

PR #366, opened 2023-11-02
[refactor] Switch from Rc<T> to Arc<T>

### LINKED ISSUE #250: Refactor: Consider switching from `Rc` to `Arc`
We're looking at doing lock-free concurrency in the future. That'd mean you'd possibly share references to `PrivateDirectory`s or `PublicFile`s across threads, so you'd need to use `Arc`s.

Enabling lock-free concurrency is kinda blocked by implementing our CRDT `merge` functions on `PrivateDirectory` and `PublicDirectory`.

Perhaps it makes sense to do this via feature-gating, so the wasm API or any single-threaded consumer can continue using `Rc`s (I know rs-ucan already does this by feature-gating on `target = wasm32-unknown-unknown` for figuring out if `async_trait` should be `?Send`).

Would also love to know exactly what this could help with & enable from folks like @Frando <3

fabricedesre:
## Summary

This PR fixes/implements the following **bugs/features**

- [X] https://github.com/wnfs-wg/rs-wnfs/issues/250

## Closing issues

closes #250 

## After Merge

- [X] Does this change require a release to be made? Is so please create and deploy the release

I didn't bump the crate version number yet. This change is breaking the API backward compatibility so that should be a major semver change, but I'll let you decide what you want to do about it.

### DIFF
--- a/wnfs-bench/hamt.rs
+++ b/wnfs-bench/hamt.rs
@@ -4,7 +4,7 @@ use criterion::{
     Criterion, Throughput,
 };
 use proptest::{arbitrary::any, collection::vec, test_runner::TestRunner};
-use std::{cmp, rc::Rc, sync::Arc};
+use std::{cmp, sync::Arc};
 use wnfs_common::{
     async_encode, decode, libipld::cbor::DagCborCodec, utils::Sampleable, BlockStore, Link,
     MemoryBlockStore,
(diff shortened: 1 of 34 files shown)


matheus23:
> I didn't bump the crate version number yet. This change is breaking the API backward compatibility so that should be a major semver change, but I'll let you decide what you want to do about it.

Yeah we're *super* not doing semver yet, but that's fine if that's another PR.

---

Seems like the wasm code needs the same `Rc` -> `Arc` treatment, but I think that should be fine. IIUC, [`Arc` in `wasm32-unknown-unknown` ends up being identical to `Rc` in functionality](https://users.rust-lang.org/t/arc-works-when-using-wasm-bindgen-for-html5/97736/5) (which is a bummer). But it should compile.

PR #368, opened 2023-11-07
WIP - Make the BlockStore sendable.

fabricedesre:

PR #372, opened 2023-11-21
feat: Make rs-wnfs work in multithreaded contexts

matheus23:
The main goal of this PR is to enable using rs-wnfs in multithreaded contexts, e.g. in axum webservers.
We have a test repo to check whether that works in an MVP: https://github.com/fabricedesre/wnfs-mtload/

Previously that wasn't possible, since the async futures from rs-wnfs weren't `Send`, so you couldn't have a multi-threaded work-stealing async runtime work on them.
The reasons they weren't sync were:
- Futures would capture an `impl BlockStore`, and it's not necessarily known to be `Send`
- Futures would capture an `impl PrivateForest` with the same problem
- Some functions would produce `LocalBoxFuture` or `LocalBoxStream`, which aren't `Send`
- We'd use `async_trait(?Send)` and `async_recursion(?Send)` which opt-in to not having `Send` bounds, since that's what we need for wasm
- Futures would capture internal WNFS data structures like `PrivateNode`, which would use `Rc` internally instead of `Arc`, see also #250 

Some of this work was already addressed in #366. This PR *should* cover the rest.

---

There's a complication with Wasm, where we're e.g. using an external type `extern "C" { type BlockStore; ... }`, which isn't `Send` or `Sync`, and as such can't ever implement a `trait BlockStore: Send + Sync`.
To fix this, we're conditionally compiling in `Send` and `Sync` bounds (and `Arc` and `Rc` and similar) based on the target (See `send_sync_poly.rs`). This is pretty much just copying what noosphere is doing: https://github.com/subconsciousnetwork/noosphere/blob/main/rust/noosphere-common/src/sync.rs

I'm hoping eventually we just fix this and thus enable multi-threaded Wasm, too. But for now this works.


matheus23:
Replaces #368

matheus23:
TODO:
- Look at remaining `async_trait(?Send)`:
  - `#[async_trait(?Send)] pub trait ExchangeKey`
  - `#[async_trait(?Send)] pub trait PrivateKey`
  - `WnfsBlockStore` and similar wrappers in Wasm code
- Fix `wnfs-wasm` crate type errors.

fabricedesre (approved):
That looks great! Do you plan to do a version bump and release in another PR?

matheus23:
> That looks great! Do you plan to do a version bump and release in another PR?

Yep. Soon-ish :v:

PR #386, opened 2023-12-20
Match the public and private APIs for content reading

icidasset:
## Summary

I've made it so that the public and private APIs for reading content better match.

`wnfs` changes:
- Made the length param optional and renamed the offset param to match the public side.
- Copied the doc test from the public side to the private side.
- Made it so that `PrivateFile.get_content` uses its `read_at` method.
- Added `PublicFile.get_content` to reflect the private side.

`wnfs-wasm` changes:
- Added `PrivateFile.read_at`
- Made it so that `PrivateFile.get_content` uses its `read_at` method.
- Added `PublicFile.get_content`
- Adjusted tests to use new and changed methods.

## Test plan (required)

Added and adjusted some tests.

## After Merge

- [ ] Does this change require a release to be made? Is so please create and deploy the release

### DIFF
--- a/wnfs-wasm/tests/private.spec.ts
+++ b/wnfs-wasm/tests/private.spec.ts
@@ -29,7 +29,7 @@ test.describe("PrivateDirectory", () => {
         new Date(),
         initialForest,
         store,
-        rng
+        rng,
       );
 
       return await rootDir.lookupNode("text.txt", true, forest, store);
@@ -56,7 +56,7 @@ test.describe("PrivateDirectory", () => {
         new Date(),
         initialForest,
         store,
-        rng
+        rng,
       );
 
       var { rootDir, forest } = await rootDir.write(
@@ -66,22 +66,12 @@ test.describe("PrivateDirectory", () => {
         new Date(),
         forest,
         store,
-        rng
+        rng,
       );
 
-      var result0 = await rootDir.getNode(
-        ["pictures", "cats", "tabby.png"],
-        true,
-        forest,
-        store
-      );
+      var result0 = await rootDir.getNode(["pictures", "cats", "tabby.png"], true, forest, store);
 
-      var result1 = await rootDir.getNode(
-        ["pictures", "dogs", "bingo.png"],
-        true,
-        forest,
-        store
-      );
+      var result1 = await rootDir.getNode(["pictures", "dogs", "bingo.png"], true, forest, store);
 
       console.log(result0);
       console.log(result1);
@@ -93,9 +83,7 @@ test.describe("PrivateDirectory", () => {
     expect(result1).toBeUndefined();
   });
 
-  test("lookupNode cannot fetch file not added to directory", async ({
-    page,
-  }) => {
+  test("lookupNode cannot fetch file not added to directory", async ({ page }) => {
     const result = await page.evaluate(async () => {
       const {
         wnfs: { PrivateDirectory, PrivateForest },
@@ -131,7 +119,7 @@ test.describe("PrivateDirectory", () => {
         new Date(),
         initialForest,
         store,
-        rng
+        rng,
       );
 
       var { rootDir, forest } = await rootDir.write(
@@ -141,15 +129,10 @@ test.describe("PrivateDirectory", () => {
         new Date(),
         forest,
         store,
-        rng
+        rng,
       );
 
-      var result = await rootDir.getNode(
-        ["pictures", "cats", "tabby.png"],
-        true,
-        forest,
-        store
-      );
+      var result = await rootDir.getNode(["pictures", "cats", "tabby.png"], true, forest, store);
 
       return result;
     });
@@ -175,7 +158,7 @@ test.describe("PrivateDirectory", () => {
         new Date(),
         initialForest,
         store,
-        rng
+        rng,
       );
 
       var { rootDir, forest } = await rootDir.write(
@@ -185,7 +168,7 @@ test.describe("PrivateDirectory", () => {
         new Date(),
         forest,
         store,
-        rng
+        rng,
       );
 
       var { result } = await rootDir.ls(["pictures"], true, forest, store);
@@ -217,7 +200,7 @@ test.describe("PrivateDirectory", () => {
         new Date(),
         initialForest,
         store,
-        rng
+        rng,
       );
 
       var { rootDir, forest } = await rootDir.write(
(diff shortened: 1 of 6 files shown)


wnfs-wasm/tests/private.spec.ts
```diff
@@ -141,15 +129,10 @@ test.describe("PrivateDirectory", () => {
...
+        rng,
       );
 
-      var result = await rootDir.getNode(
-        ["pictures", "cats", "tabby.png"],
-        true,
-        forest,
-        store
-      );
+      var result = await rootDir.getNode(["pictures", "cats", "tabby.png"], true, forest, store);
```
icidasset, line 135: Just now noticed that my editor the auto formatting thing again 🤦 Let me know if I should undo this.

wnfs/src/private/file.rs
```diff
@@ -504,17 +504,66 @@ impl PrivateFile {
...
+    ///     assert_eq!(content, b"Hello, World!\nHello, World!\n");
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
     pub async fn read_at<'a>(
         &'a self,
-        offset: usize,
-        len: usize,
+        byte_offset: usize,
```
matheus23, line 552: While you're here, is it fine if I ask you to change `byte_offset: usize` to `byte_offset: u64`, similar to what's in the public side?
It's a tiny bit of a refactor. The `u64` should carry through to the `.stream` call (`block_index` should also be a `u64`), the very leaf of the call tree being `create_block_name`, which today turns the `usize` block index into a `u64`, but should instead just take a `u64`.

(And `len`s and `len_limit`s generally should keep being `usize`s)

wnfs-wasm/tests/private.spec.ts
matheus23: It's fine. Don't worry about it :v:

matheus23 (approved):
I'd take this as-is, but while "we have the hood up" and working towards public & private APIs being aligned, would you be up for tackling a small refactor from `usize` to `u64` for byte offsets? (See the comment)

wnfs/src/private/file.rs
icidasset: Sure thing. One thing that's a bit annoying here is that a lot of the other numbers we're working with are `usize`s. Do we cast those usizes to u64, or the other way around?

matheus23: for block indices, work in u64 for as long as possible. We should only need to cast that to `usize` to work with native array indices.

icidasset: Ok, updated. Not sure if this was the right approach, feels a bit messy.

wnfs/src/private/file.rs
```diff
@@ -921,35 +971,44 @@ impl PrivateForestContent {
...
         store: &'a impl BlockStore,
     ) -> Result<Vec<u8>> {
-        let block_content_size = MAX_BLOCK_CONTENT_SIZE;
-        let chunk_size_upper_bound = (self.get_size_upper_bound() - offset).min(len);
+        let block_content_size = MAX_BLOCK_CONTENT_SIZE as u64;
+        let mut chunk_size_upper_bound = self.get_size_upper_bound() - byte_offset as usize;
+
+        if len_limit.is_some() {
+            chunk_size_upper_bound = chunk_size_upper_bound.min(len_limit.unwrap());
+        }
```
matheus23, line 984:
```suggestion
        if let Some(len_limit) = len_limit {
            chunk_size_upper_bound = chunk_size_upper_bound.min(len_limit);
        }
```
This gets rid of a call to a possibly panicking `.unwrap()` :v:

wnfs/src/private/file.rs
```diff
@@ -1307,7 +1366,10 @@ mod proptests {
                 .await
                 .unwrap();
             disk_file.read_exact(&mut source_content).await.unwrap();
-            let wnfs_content = file.read_at(offset, size, forest, store).await.unwrap();
+            let wnfs_content = file
+                .read_at(offset as u64, Some(size), forest, store)
```
matheus23, line 1370: `offset` should probably just be generated as `u64`. So in the argument of the function:
`#[strategy(0..FIXTURE_SCHERZO_SIZE as u64)] offset: u64,`

wnfs/src/private/file.rs
```diff
@@ -921,35 +971,44 @@ impl PrivateForestContent {
...
-                (offset - index * block_content_size).min(chunk.len())
+                (byte_offset - index * block_content_size).min(chunk.len() as u64)
             } else {
                 0
             };
-            let to = if index == last_block {
-                (offset + len - index * block_content_size).min(chunk.len())
+            let to = if Some(index) == last_block {
+                (byte_offset + len_limit.unwrap() as u64 - index * block_content_size)
+                    .min(chunk.len() as u64)
```
matheus23, line 1006: I see how this `.unwrap()` is valid, but it's not super easy to figure out why. (It's because `Some(index) == last_block` is only true, when `last_block` is `Some(_)`, which is only true if `len_limit` is `Some(_)`.)
```suggestion
            let to = if let Some(len_limit) = len_limit && Some(index) == last_block {
                (byte_offset + len_limit as u64 - index * block_content_size)
                    .min(chunk.len() as u64)
```

wnfs/src/private/file.rs
```diff
@@ -875,7 +925,7 @@ impl PrivateForestContent {
         Ok(PrivateForestContent {
             key,
             base_name: forest.get_accumulated_name(&base_name),
-            block_count: block_index,
+            block_count: block_index as usize,
```
matheus23, line 928: `Can` you try changing `block_count: u64`? That may make stuff easier.

icidasset: Yeah I tried that, requires a bunch more changes to various types 🤔

matheus23: Ah okay. Then we'll leave it at that.

wnfs/src/private/file.rs
icidasset: Yeah makes sense! Unfortunately this gives me an error, says `let` expressions in this position are unstable 👀

matheus23: What version of rustc do you use?

matheus23: Lol nevermind I see it in CI

wnfs/src/private/file.rs
icidasset: Done 👍

wnfs/src/private/file.rs
matheus23: Well, then let's just make it `.unwrap_or_default()` for now:
```rs
            let to = if Some(index) == last_block {
                (byte_offset + len_limit.unwrap_or_default() as u64 - index * block_content_size)
                    .min(chunk.len() as u64)
```

wnfs/src/private/file.rs
matheus23: It's okay. The most important piece is getting the API to support very large files & to align public and private APIs.
We can do a cleanup round later :v:

PR #438, opened 2024-06-22
refactor: Switch to `ipld-core` and eigerco/blockstore

matheus23:
- Switches to the `blockstore` crate for better reuse across the ecosystem
- Switches away from `libipld`/`libipld_core` in favor of `ipld-core`, the new successor. This also enables depending on the new `cid` crate version `0.11` instead of `0.10`.

TODO:
- [ ] Investigate snapshot failures. There shouldn't be any data format changes at all.
- [ ] Continue/finish the refactor in `wnfs-wasm`
- [ ] Figure out what to do with `Blake3Block`. With the previous blockstore, rs-wnfs didn't care which hash function you used. The blockstore chooses. But now the blockstore expects the outside to choose. This is annoying.


matheus23:
The last point was a deal-breaking somewhat. All of the rest should now happen in #460

PR #464, opened 2026-06-15
fix: get rid of yanked core2 crate by updating hamt dependencies

klpx:
Hi @matheus23 ! This is very interesting project i'm digging into. However it's not possible at the moment to use `wnfs-hamt` crate, because it's indirect dependency - `core2` - has been yanked (https://rustsec.org/advisories/RUSTSEC-2026-0105).

All other crate in this repository already uses core2-free versions of multihash and serde_ipld_dagcbor, so this PR could be treated as aligning wnfs-hamt deps with the siblings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,27 +3,12 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.25.1"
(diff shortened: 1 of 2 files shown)


matheus23 (approved):

matheus23:
Thanks!

